PR #4, opened 2025-12-14
Feat/openblt

### LINKED ISSUE #1: Implement OpenBLT Bootloader
# Description

Since the cockpit firmware will use the canlib it is needed to update and upload the firmware via CAN bus.
To achieve this the openblt bootloader needs to be implemented and tested into the project.

## Implementation Steps

1. Get openblt files of the specific microcontroller family
2. Integrate the openblt bootloader inside the project by creating the `platformio.ini`
3. Test that the bootloader is working correctly (to test the board ois needed)

Bridiro:
# Purpose

This PR's purpose is to add OpenBLT bootloader.

# Overview

Linker files has been modified to accomodate memory constraints. Also `libraster-sw` has been added as a dependency to the bootloader to display a boot screen, for both information and debug. Framebuffer and display peripheral have been setup as far as possible without testing hardware, so it will need changes in the future.

If not known, on dual core architectures only CM7 is needed to act as a bootloader, so CM4 will actually be disabled from booting automatically. In fact all the dual core syncing code has been disabled and `platformio.ini` only builds CM7 code.

From CubeMX I enabled `DMA2D` and `JPEG_Decoder` for the possibility of using them to display more complex interfaces while booting.

# Guidance

As only files has been copied from Feaser repository, the only file worth checking is probably `but_conf.h` (`CM/Core/Inc/`).

This PR may close #1


steering-wheel-bootloader-sw/CM7/Core/Inc/blt_conf.h
```diff
@@ -0,0 +1,83 @@
...
+#define BOOT_COM_NET_DEFERRED_INIT_ENABLE (0)
+
+/* USB disabled */
+#define BOOT_COM_USB_ENABLE (0)
+#define BOOT_COM_USB_TX_MAX_DATA (0)
+#define BOOT_COM_USB_RX_MAX_DATA (0)
+
+/* CAN enabled as the ONLY interface */
+#define BOOT_COM_CAN_ENABLE (1)
+#define BOOT_COM_CAN_BAUDRATE (500000)
```
Tonidotpy: At the moment we are using 1M for the CAN baudrate, for the next car we are probably going to use the same

Tonidotpy (approved):
- CAN Baudrate can be changed later
- Add bootloader compilation step in GitHub workflows

PR #10, opened 2026-02-17
Feat/inputs

### LINKED ISSUE #7: Implement Input module(s)
# Description

All inputs are handled using interrupts on the M4 core. The M7 core has to react to those inputs prompting the driver on the screen.
To implement the module we have to find a way to store and communicate inputs between cores.
Of course the module has to reflect every possible input and every function associated with it.

This will handle both buttons and encoders while being as hardware-agnostic as possible.

## Implementation Steps

1. write down every possible function of every button
2. start implementing from core M4 as it's the main focus of this module
3. find a way in which the core M7 could receive meaningful input datas, while keeping memory usage low
4. create tests for both cores' modules

## Dependencies

None

## Additional Notes

None

Bridiro:
# Purpose

The idea of this PR is to  add the foundations of the input module, from handling the event in the CM4 to transmitting it to the CM7 to properly display it.

By opening the pull request so early I'm searching for validation in my thought process as I'm not sure if it's liked or not.

This PR may close #7 

# Overview

I added code in both cores.
- **CM4**: main changes, here I added a system where a function is called by the EXTI handler, passing the PIN that triggered it to correctly handle it, and to call a callback provided by the user, that notify the CM7.
- **CM7**: here it's simpler as based on what comes the module simply calls a callback based on the event type. 

# Guidance

Just check out the 2 modules. I suggest to completely skip the "drivers" part as of right now there are only mock functions.


Common/Inc/steering/inputs-shared.h
Tonidotpy, line 1: - Remember to prefix all enums fields with the enum name. (e.g. InputEventType -> INPUT_EVENT_TYPE_KNOB, KnobID -> KNOB_ID_FRONT_LEFT, etc...)
- Also I'm not really sure if it is practical to keep all shared definitions into a single file instead of separating them into different files. For example all input related stuff may be moved into `Common/shared-input.h` or something like that

CM7/Core/Inc/steering/input-events-driver.h
```diff
@@ -0,0 +1,14 @@
+#ifndef INPUTS_EVENTS_DRIVER_H
+#define INPUTS_EVENTS_DRIVER_H
+
+#include "input-events.h"
+
+extern struct InputEventHandler input_event_handler;
+extern struct InputEvent shared_data;
```
Tonidotpy: Avoid the use of extern variables!!!
They increase code complexity and make code difficult to read and maintain as well as sometimes causing unwanted behaviors.

CM4/Core/Src/inputs/inputs.c
```diff
@@ -0,0 +1,131 @@
...
+ * \brief Hardware-agnostic input handling implementation for the steering wheel.
+ */
+
+#include "inputs.h"
+#include <string.h>
+
+enum InputsReturnCode inputs_init(
+    struct InputHandler *handler,
+    input_event_notify_callback callback) {
+    if (!handler || !callback) {
```
Tonidotpy: Compare against `NULL` instead of using the _not_ operator. It increases code readability and maintainability.

CM4/Core/Src/inputs/inputs.c
```diff
@@ -0,0 +1,131 @@
...
+        return INPUTS_ERROR;
+    }
+
+    memset(handler, 0, sizeof(struct InputHandler));
```
Tonidotpy: Using directly the variable instead of the type in the `sizeof` operator is easier to maintain since if the variable type changes this line don't. Also `sizeof` is evaluated at compile time so there is no pointer dereferencing thus no bad memory access can happen.
```suggestion
    memset(handler, 0, sizeof(*handler));
```

CM4/Core/Inc/steering/inputs-driver.h
```diff
@@ -0,0 +1,12 @@
+#ifndef INPUTS_DRIVER_H
+#define INPUTS_DRIVER_H
+
+#include "shared.h"
+#include "inputs.h"
+
+extern struct InputEvent shared_data;
+extern struct InputHandler input_handler;
```
Tonidotpy: `extern` should not be used!!!

CM4/Core/Src/steering/inputs-driver.c
```diff
@@ -0,0 +1,12 @@
+#include "inputs-driver.h"
+
+struct InputEvent shared_data
+    __attribute__((section(".shared_axi"), aligned(32)));
+
+struct InputHandler input_handler = { 0 };
```
Tonidotpy: Should this be defined here or somewhere else?
Is it specific of the input drivers or is more abstract (thus for the input module)?

CM4/Core/Inc/inputs/inputs.h
```diff
@@ -0,0 +1,129 @@
...
+ */
+
+#ifndef INPUTS_H
+#define INPUTS_H
+
+#include "shared.h"
+#include <stdint.h>
+#include <stdbool.h>
+
+#define LONG_PRESS_THRESHOLD_MS 500
```
Tonidotpy: Also, should this be here?
```suggestion
#define INPUTS_LONG_PRESS_THRESHOLD_MS (500)
```

CM4/Core/Inc/steering/inputs.h
Tonidotpy, line 1: As for the shared file remember to prefix everything using the name of the module/file

CM4/Core/Inc/inputs/inputs.h
```diff
@@ -0,0 +1,129 @@
...
+enum ButtonState {
+    BUTTON_STATE_IDLE,         /*!< Button is not pressed */
+    BUTTON_STATE_PRESSED,      /*!< Button is currently pressed */
+    BUTTON_STATE_LONG_PRESSED, /*!< Button has been held long enough to be considered a long press */
+};
+
+/*!
+ * \brief Button tracking structure
+ */
+struct ButtonTracker {
```
Tonidotpy: `ButtonHandler` may be bettere as a name than `ButtonTracker`, you are not following the button you are handling its state.

CM4/Core/Inc/inputs/inputs.h
```diff
@@ -0,0 +1,129 @@
...
+    enum ButtonState state; /*!< Current state of the button */
+    uint32_t press_tick;    /*!< Tick count when the button was pressed */
+    bool enabled;           /*!< Whether this button is enabled for tracking */
+};
+
+/*!
+ * \brief Knob (encoder) tracking structure
+ */
+struct KnobTracker {
```
Tonidotpy: `KnobHandler` may be bettere as a name than `KnobTracker`, you are not following the knob you are handling its state.

CM4/Core/Inc/inputs/inputs.h
```diff
@@ -0,0 +1,129 @@
...
+ * \param knob_id Identifier of the knob that changed state
+ * \param current_position Current position of the encoder
+ *
+ * \retval INPUTS_OK if the knob state was updated successfully
+ * \retval INPUTS_ERROR if there was an error updating the knob state
+ */
+enum InputsReturnCode inputs_check_knobs(
+    struct InputHandler *handler,
+    enum KnobID knob_id,
+    int16_t current_position);
```
Tonidotpy: - Why this inconsistency in naming if the two functions both update inputs?
- Also, `check` is ambiguous as a word in this context
- You also have a common structure for both inputs, can this two functions be merged into a single one? Choose the simpler option between a single more abstract function and two or more separate more specific functions

CM4/Core/Inc/steering/inputs.h
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * \param handler Pointer to the input handler structure
+ * \param current_tick_ms Current tick count in milliseconds
+ *
+ * \retval INPUTS_OK if the update was successful
+ * \retval INPUTS_ERROR if there was an error during the update
+ */
+enum InputsReturnCode inputs_update(
+    struct InputHandler *handler,
+    uint32_t current_tick_ms);
```
Tonidotpy: Not sure what exactly this function is updating, the name and description are ambiguous and not clear enough

CM4/Core/Src/inputs/inputs.c
```diff
@@ -0,0 +1,131 @@
...
+
+    return INPUTS_OK;
+}
+
+enum InputsReturnCode inputs_on_button_interrupt(
+    struct InputHandler *handler,
+    enum ButtonID button_id,
+    bool pressed,
+    uint32_t current_tick_ms) {
+    if (!handler || button_id < 0 || button_id >= BUTTON_COUNT) {
```
Tonidotpy:
```suggestion
    if (handler == NULL || button_id < 0 || button_id >= BUTTON_COUNT) {
```

CM4/Core/Src/steering/inputs-api.c
```diff
@@ -0,0 +1,131 @@
...
+
+    if (pressed) {
+        if (btn->state == BUTTON_STATE_IDLE) {
+            btn->state = BUTTON_STATE_PRESSED;
+            btn->press_tick = current_tick_ms;
+
+            event.type = INPUT_EVENT_TYPE_BUTTON_PRESS;
+            return handler->notify_callback(event);
+        }
+    } else if (btn->state == BUTTON_STATE_PRESSED || btn->state == BUTTON_STATE_LONG_PRESSED) {
```
Tonidotpy: Not sure how the long press is handled, is the logic missing?

CM4/Core/Src/inputs/inputs.c
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    return INPUTS_OK;
+}
+
+enum InputsReturnCode inputs_check_knobs(
+    struct InputHandler *handler,
+    enum KnobID knob_id,
+    int16_t current_position) {
+    if (!handler || knob_id < 0 || knob_id >= KNOB_COUNT) {
```
Tonidotpy:
```suggestion
    if (handler == NULL || knob_id < 0 || knob_id >= KNOB_COUNT) {
```

CM4/Core/Src/inputs/inputs.c
```diff
@@ -0,0 +1,131 @@
...
+
+    int16_t delta = current_position - knob->last_position;
+
+    knob->last_position = current_position;
+
+    if (delta != 0) {
+        struct InputEvent event;
+        event.type = INPUT_EVENT_TYPE_KNOB;
+        event.knob.knob_id = knob_id;
+        event.knob.delta = (int8_t)delta;
```
Tonidotpy: Structs supports direct initialization, every other field is initialized to 0 if I remember correctly.
As it is for now additional fields are not initialized and if used later they may cause undefined behavior
```suggestion
        struct InputEvent event = {
            .type = INPUT_EVENT_TYPE_KNOB,
            .knob.knob_id = knob_id,
            .knob.delta = (int8_t)delta
        };
```

CM4/Core/Src/inputs/inputs.c
```diff
@@ -0,0 +1,131 @@
...
+
+        return handler->notify_callback(event);
+    }
+
+    return INPUTS_OK;
+}
+
+enum InputsReturnCode inputs_update(struct InputHandler *handler, uint32_t current_tick_ms) {
+    if (!handler) {
```
Tonidotpy:
```suggestion
    if (handler == NULL) {
```

CM4/Core/Src/inputs/inputs.c
```diff
@@ -0,0 +1,131 @@
...
+        if (!btn->enabled) {
+            continue;
+        }
+
+        if (btn->state == BUTTON_STATE_PRESSED && current_tick_ms - btn->press_tick >= LONG_PRESS_THRESHOLD_MS) {
+            btn->state = BUTTON_STATE_LONG_PRESSED;
+
+            struct InputEvent event;
+            event.type = INPUT_EVENT_TYPE_BUTTON_LONG_PRESS;
+            event.button.button_id = i;
```
Tonidotpy: Same as above:
```suggestion
            struct InputEvent event = {
                .type = INPUT_EVENT_TYPE_BUTTON_LONG_PRESS,
                .button.button_id = i
            };
```

CM4/Core/Src/inputs/inputs.c
```diff
@@ -0,0 +1,131 @@
...
+
+    return INPUTS_OK;
+}
+
+enum InputsReturnCode inputs_update(struct InputHandler *handler, uint32_t current_tick_ms) {
+    if (!handler) {
+        return INPUTS_ERROR;
+    }
+
+    for (uint8_t i = 0; i < BUTTON_COUNT; i++) {
```
Tonidotpy: Better not to use `uint8_t` inside for loops, prefer `size_t` or other common integer types.
Talking by experience here, this can easily cause some strange bugs in the code 🫤 and you are not saving that much memory anyway so it is not that convenient.
Also not sure how it plays with alignment so it may also be slower but that might be a stupid statement.

CM7/Core/Src/steering/input-events-driver.c
```diff
@@ -0,0 +1,22 @@
+#include "input-events-driver.h"
+#include "shared.h"
+
+struct InputEventHandler input_event_handler = { 0 };
+struct InputEvent shared_data
+    __attribute__((section(".shared_axi"), aligned(32)));
```
Tonidotpy: Are you sure it goes in here and not in the `input-event` module/file or somewhere else?

CM7/Core/Inc/steering/input-events.h
```diff
@@ -0,0 +1,102 @@
...
+#include "shared.h"
+#include <stdint.h>
+
+/*!
+ * \brief Return codes for input event callbacks.
+ */
+enum InputEventReturnCode {
+    INPUT_EVENT_OK,    /*!< Event handled successfully. */
+    INPUT_EVENT_ERROR, /*!< An error occurred while handling the event. */
+};
```
Tonidotpy: Missing `RC` in prefix:
```suggestion
enum InputEventReturnCode {
    INPUT_EVENT_RC_OK,    /*!< Event handled successfully. */
    INPUT_EVENT_RC_ERROR, /*!< An error occurred while handling the event. */
};
```

Tonidotpy (changes requested):
- I don't really understand the architecture for the input modules. What do you mean by driver? Because we may be interpreting the word driver into two different manners
- Remember to use `libeagletrt` definitions and functionalities whenever possible
- Remember to first lay out the architecture design and then implementing it, otherwise it will be difficult to have a simple, efficient and clear implementation of the software. This is especially important if you are implementing an **event-based architecture**, I suggest to look some resources online before starting right on with the implementation

Tonidotpy:
@Bridiro Avoid shortening the microcontroller pin names, they are becoming incomprehensible. Always prefer readable over short names.

Bridiro:
> @Bridiro Avoid shortening the microcontroller pin names, they are becoming incomprehensible. Always prefer readable over short names.

@Tonidotpy It is actually done to have the same naming convention as the hardware team. This is to prevent issues that we found the last year during my training on the old steering wheel where naming was different and nothing matched. I think names are pretty readable and if extended, they would be too long to write (like shortening `ReturnCode` to `RC`, you just know what it is).

Tonidotpy:
> > @Bridiro Avoid shortening the microcontroller pin names, they are becoming incomprehensible. Always prefer readable over short names.
> 
> @Tonidotpy It is actually done to have the same naming convention as the hardware team. This is to prevent issues that we found the last year during my training on the old steering wheel where naming was different and nothing matched. I think names are pretty readable and if extended, they would be too long to write (like shortening `ReturnCode` to `RC`, you just know what it is).

It's good to keep the same names between the kicad project and the cubemx project but the same reasoning applies also for the hardware area.
Shortening `ReturnCode` was agreed by everyone because it is used very often almost everywhere, but that does not mean it should be done everywhere.
Anyway, how can `ENC_SL_SW` be readable? Does ENC stand for encoder or enclosure? What does SL mean ShiftLeft, SunLight, SLip?
Also complaining that names are too long too write seem just like a **skill issue** to me, especially nowadays with auto completion and copilot features.

**Readability >> Conciseness**
Think about a new member that has never programmed anything inside an embedded system...

Bridiro:
> > > @Bridiro Avoid shortening the microcontroller pin names, they are becoming incomprehensible. Always prefer readable over short names.
> > 
> > 
> > @Tonidotpy It is actually done to have the same naming convention as the hardware team. This is to prevent issues that we found the last year during my training on the old steering wheel where naming was different and nothing matched. I think names are pretty readable and if extended, they would be too long to write (like shortening `ReturnCode` to `RC`, you just know what it is).
> 
> It's good to keep the same names between the kicad project and the cubemx project but the same reasoning applies also for the hardware area. Shortening `ReturnCode` was agreed by everyone because it is used very often almost everywhere, but that does not mean it should be done everywhere. Anyway, how can `ENC_SL_SW` be readable? Does ENC stand for encoder or enclosure? What does SL mean ShiftLeft, SunLight, SLip? Also complaining that names are too long too write seem just like a **skill issue** to me, especially nowadays with auto completion and copilot features.
> 
> **Readability >> Conciseness** Think about a new member that has never programmed anything inside an embedded system...

Then I'll try to talk to Angelo but he'll kill me if I ask to change every name

CM7/Core/Src/ltdc.c
```diff
@@ -21,6 +21,13 @@
 #include "ltdc.h"
 
 /* USER CODE BEGIN 0 */
+#include "eagletrt-api.h"
+
+EAGLETRT_STATIC uint32_t framebuffer1[800 * 480]
+    __attribute__((section(".framebuffer"), aligned(32)));
+
+EAGLETRT_STATIC uint32_t framebuffer2[800 * 480]
+    __attribute__((section(".framebuffer"), aligned(32)));
```
Tonidotpy: Add macros to define frame buffer width (800) and height (480)

Common/Inc/inputs-shared.h
```diff
@@ -0,0 +1,97 @@
...
+
+/*!
+ * \brief HSEM ID used by CM4 to notify CM7 about input events.
+ */
+#define HSEM_INPUT_ID (1U)
+
+/*!
+ * \brief Size of the IPC input event ring buffer.
+ */
+#define IPC_INPUT_QUEUE_SIZE 64
```
Tonidotpy:
```suggestion
#define IPC_INPUT_QUEUE_SIZE (64U)
```

Common/Inc/steering/inputs-shared.h
Tonidotpy, line 1: Every identifier of this module is missing its prefix.
You may change the module name to have shorter identifier names but I **strongly** suggest you to put it as prefix for each identifier to avoid possible conflicts even if the probability is pretty low.
In my opinion you can split this module into two: **IPC** used for communication between the two cores, and **peripherals** (or a better name that resemble the steering wheel physical inputs) for handling of button, knobs, etc....

Common/Inc/steering/inputs-shared.h
```diff
@@ -0,0 +1,97 @@
...
+/*!
+ * \brief Size of the IPC input event ring buffer.
+ */
+#define IPC_INPUT_QUEUE_SIZE 64
+
+/*!
+ * \brief Enumeration of input event types
+ */
+enum InputEventType {
+    INPUT_EVENT_TYPE_KNOB,
```
Tonidotpy: What type of event is knob?

Common/Inc/inputs-shared.h
```diff
@@ -0,0 +1,97 @@
...
+    };
+};
+
+/*!
+ * \brief Lock-free single-producer single-consumer ring buffer for input events.
+ */
+struct IPCInputQueue {
+    uint32_t write_idx;                             /*!< Next slot to write (owned by CM4) */
+    uint32_t read_idx;                              /*!< Next slot to read  (owned by CM7) */
+    struct InputEvent events[IPC_INPUT_QUEUE_SIZE]; /*!< Ring buffer storage */
```
Tonidotpy: A Ring Buffer library already exists, you can use that instead of implementing your own.
It also allows to define callback to handle critical sections which are called direclty inside the library to minimize overhead.
For now the Arena Allocator used inside the Ring Buffer library does not support the use of a specific memory area or section (there is an issue for that), this can be implemented on a new version.
It can also be implemented inside the Ring Buffer a `ring_buffer_init_unmanaged` function that does not use the arena but instead it accepts a pointer to an external array (this second option is more likely to create bugs if memory is not managed correctly by the user of the library).

Common/Src/steering/ipc-api.c
Tonidotpy, line 1: Yet another Ring Buffer implementation...

Common/Inc/inputs-shared.h
```diff
@@ -0,0 +1,97 @@
...
+
+#ifndef INPUTS_SHARED_H
+#define INPUTS_SHARED_H
+
+#include <stdint.h>
+
+/*!
+ * \brief HSEM ID used by CM4 to notify CM7 about input events.
+ */
+#define HSEM_INPUT_ID (1U)
```
Tonidotpy: This seems more related to hardware dependant code, it should not be defined here.
If needed inside the `Common` folder you can create a new header file outside of the `steering` folder to indicate that it may contain hardware dependant code.

CM4/Core/Inc/steering/inputs.h
Tonidotpy, line 1: - Move function declarations to `inputs-api.h`
- It seem like the input definitions are mixed between this module and the one in the `Common` folder.

CM4/Core/Inc/steering/inputs.h
```diff
@@ -0,0 +1,135 @@
...
+enum InputsReturnCode {
+    INPUTS_RC_OK,           /*!< Operation successful */
+    INPUTS_RC_NOTIFY_ERROR, /*!< Error while notifying */
+    INPUTS_RC_ERROR,        /*!< General error */
+};
+
+/*!
+ * \brief Button states for tracking long presses
+ */
+enum ButtonState {
```
Tonidotpy: Why are `ButtonId` and `KnobId` inside common but the `ButtonState` not?

CM4/Core/Src/steering/inputs.c
```diff
@@ -0,0 +1,157 @@
...
+#include "eagletrt.h"
+
+/*!
+ * \brief Dispatch an event to both the notify and action callbacks.
+ *
+ * The notify callback is always called first.
+ * If either fails the first non-OK code is returned, but both
+ * are always attempted.
+ */
+EAGLETRT_STATIC enum InputsReturnCode inputs_dispatch(
```
Tonidotpy:
```suggestion
EAGLETRT_STATIC enum InputsReturnCode prv_inputs_dispatch(
```

CM7/Core/Src/steering/input-events-api.c
```diff
@@ -0,0 +1,51 @@
...
+}
+
+void input_events_api_handle_event(struct InputEvent *event) {
+    if (event == NULL) {
+        return;
+    }
+
+    switch (event->type) {
+        case INPUT_EVENT_TYPE_BUTTON_PRESS:
+            if (handler.on_button_press) {
```
Tonidotpy: Always compare against `NULL`, same for the below cases.
```suggestion
            if (handler.on_button_press != NULL) {
```

Tonidotpy (changes requested):
- Remember to also explain the enum field using doxygen comments.
- There should be a `steering` folder also inside the `Common` directory for both `Inc` and `Src`
- Some identifier are missing the module name as prefix.

CM4/Core/Inc/steering/inputs.h
Bridiro: this is because `ButtonState` is used to keep track of the "position" (state) of the button (which does not concern CM7, so does not need to be shared), while `ButtonID` and `KnobID` need to be shared with CM7 to send events. This lets CM7 know on which button the event has happened. I find this the best solution to share only what's really needed.

Tonidotpy: In my opinion it is better to organize the code logically with each expression/function in its specific module and if necessary "share" the entire module instead of the signel expression/function.
Do as you think it's best.

CM7/Core/Src/main.c
```diff
@@ -60,22 +62,20 @@
...
-
 /* USER CODE END PV */
 
 /* Private function prototypes -----------------------------------------------*/
 void SystemClock_Config(void);
 static void MPU_Config(void);
 /* USER CODE BEGIN PFP */
 
+void HAL_HSEM_FreeCallback(uint32_t SemMask) {
+    if (SemMask & (1 << HSEM_INPUT_ID)) {
```
Tonidotpy: There is no error here even if the expression may return a value greater than 1 (it works anyway because it is still treated as true), but it may be better and clearer to use the `EAGLETRT_API_BIT_GET` macro provided by our main library.

CM4/Core/Inc/steering/inputs-api.h
```diff
@@ -0,0 +1,76 @@
...
+ * \param dmb_callback Callback to perform in critical section
+ * \param notify_callback Callback function to notify CM7 about input events
+ * \param action_callback Callback for local actions on input
+ *
+ * \retval INPUTS_RC_OK if initialization was successful
+ * \retval INPUTS_RC_ERROR if there was an error during initialization
+ */
+enum InputsReturnCode inputs_init(
+    struct InputsHandler *handler,
+    void (*dmb_callback)(void),
```
Tonidotpy: What does dmb stands for, is it Data Memory Barrier? Name is not clear.
I know there is the comment above but I should at least be able to tell what the callback does from its name, for additional details comments should be used.

platformio.ini
```diff
@@ -41,13 +43,14 @@ build_src_filter =
     -<Middlewares/Third_Party/FatFs/src/option/cc*.c>
     -<CM4/**>
     -<Makefile/CM4/**>
+lib_deps = https://github.com/eagletrt/libeagletrt-sw.git
```
Tonidotpy: Remember to always specify the version, otherwise any update to the dependy may brake this library.
```suggestion
lib_deps = https://github.com/eagletrt/libeagletrt-sw.git#v0.1.0
```

platformio.ini
```diff
@@ -80,13 +85,53 @@ build_src_filter =
     +<Makefile/CM4/*.s>
     -<CM7/**>
     -<Makefile/CM7/**>
+lib_deps = https://github.com/eagletrt/libeagletrt-sw.git
```
Tonidotpy: Remember to always specify the version, otherwise any update to the dependy may brake this library.
```suggestion
lib_deps = https://github.com/eagletrt/libeagletrt-sw.git#v0.1.0
```

platformio.ini
```diff
@@ -80,13 +85,53 @@ build_src_filter =
...
+platform = native
+test_framework = unity
+test_build_src = true
+build_flags = 
+    -ICM7/Core/Inc/steering
+    -Itest/include
+build_src_filter =
+    +<CM7/Core/Src/steering>
+test_filter = cm7/*
+lib_deps = https://github.com/eagletrt/libeagletrt-sw.git
```
Tonidotpy:
```suggestion
lib_deps = https://github.com/eagletrt/libeagletrt-sw.git#v0.1.0
```

platformio.ini
```diff
@@ -80,13 +85,53 @@ build_src_filter =
...
+platform = native
+test_framework = unity
+test_build_src = true
+build_flags = 
+    -ICM4/Core/Inc/steering
+    -Itest/include
+build_src_filter =
+    +<CM4/Core/Src/steering>
+test_filter = cm4/*
+lib_deps = https://github.com/eagletrt/libeagletrt-sw.git
```
Tonidotpy:
```suggestion
lib_deps = https://github.com/eagletrt/libeagletrt-sw.git#v0.1.0
```

platformio.ini
```diff
@@ -80,13 +85,53 @@ build_src_filter =
...
+platform = native
+test_framework = unity
+test_build_src = true
+build_flags = 
+    -ICommon/Inc/steering
+    -Itest/include
+build_src_filter =
+    +<Common/Src/steering>
+test_filter = common/*
+lib_deps = https://github.com/eagletrt/libeagletrt-sw.git
```
Tonidotpy:
```suggestion
lib_deps = https://github.com/eagletrt/libeagletrt-sw.git#v0.1.0
```

Common/Inc/steering/ipc-api.h
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * \param callback Callback function to handle each input event.
+ */
+void ipc_api_read_and_process_all(void (*callback)(struct InputsSharedEvent *ev));
+
+/*!
+ * \brief Push an input event to the IPC input queue.
+ *
+ * \param ev Input event to push.
+ * \param dmb_callback Callback function to ensure data memory ensure
```
Tonidotpy: What does this comment mean? You also forgot to rename this parameter as you have already done in the file `inputs-api.h`.

test/common/test_ipc/test_ipc.c
```diff
@@ -0,0 +1,64 @@
...
+ * \authors Alessandro Bridi [ale.bridi15@gmail.com]
+ *
+ * \brief Test IPC (Inter-Process Communication) functionality.
+ */
+
+#include "inputs-shared.h"
+#include "ipc-api.h"
+#include "unity.h"
+
+int called_count = 0;
```
Tonidotpy: You can use the `setUp` function to set the variable `called_count` to 0 since this function is automatically called by Unity before each test.
Otherwise you can use `tearDown` which is called at the end.

test/common/test_ipc/test_ipc.c
```diff
@@ -0,0 +1,64 @@
...
+    TEST_ASSERT_EQUAL_INT_MESSAGE(last_event.button.button_id, ev.button.button_id, "Button ID mismatch");
+    called_count = 0;
+}
+
+int main() {
+    UNITY_BEGIN();
+    RUN_TEST(test_ipc_input_queue_read_and_process_all_empty);
+    RUN_TEST(test_ipc_input_queue_read_and_process_all_non_empty);
+    RUN_TEST(test_ipc_input_queue_read_and_process_all_multiple);
+    RUN_TEST(test_ipc_input_queue_read_and_process_read_event);
```
Tonidotpy, line 127: Only 4 tests? Make sure to cover most of the possible outcomes of the various functions not only with data that you expect to have but also other possible edge cases.
You are also not testing the ipc push event logic.

Tonidotpy (changes requested):
Overall ok, a bit of work should be done for the unit tests and some small fixes in the `platformio.ini`.

CM4/Core/Inc/gpio.h
```diff
@@ -34,6 +34,8 @@ extern "C" {
 
 /* USER CODE BEGIN Private defines */
 
+extern struct InputsHandler input_handler;
```
Tonidotpy, line 37: Not sure why it is needed this extern and why is the input_handler defined inside `gpio.c`

test/cm7/test_input_events/test_input_events.c
```diff
@@ -0,0 +1,128 @@
...
+    RESET_FAKE(test_knob_rotation_callback);
+    FFF_RESET_HISTORY();
+}
+
+/*!
+ * \defgroup input_events_api_init Input Events API Initialization Tests
+ * @{
+ */
+
+void test_input_events_api_init_should_store_callbacks(void) {
```
Tonidotpy, line 25: You are missing the actual check for the callbacks. You are only looking at the return code.

test/common/test_ipc/test_ipc.c
```diff
@@ -0,0 +1,136 @@
...
+
+    RUN_TEST(test_ipc_input_queue_push_event_success);
+    RUN_TEST(test_ipc_input_queue_push_event_full);
+
+    RUN_TEST(test_ipc_input_queue_producer_consumer);
+    RUN_TEST(test_ipc_input_queue_producer_consumer_wrap);
```
Tonidotpy, line 133: Not the best names for test functions, what those function are testing is not really clear.

Tonidotpy (approved):
I don't want to slow down development too much so I'll approve the PR, but there are some improvements that may be done for the unit tests. We will discuss about them later.

Bridiro:
> I don't want to slow down development too much so I'll approve the PR, but there are some improvements that may be done for the unit tests. We will discuss about them later.

Then I'll merge and fix your suggestions in the next branch (they are only small fixes). As for the `extern`, I didn't want to create a separate **API** to get the handler, but I think that for the best, I'll instantiate the `input_handler` inside `inputs-api.c` as `static`, but I need to find a clean way to access it cleanly for the tests.

PR #11, opened 2026-03-21
Feat/led module + WS2812B driver

### LINKED ISSUE #3: Implement WS2812B driver
# Description

As the board uses WS2812B addressable leds, a driver is needed to use them. The driver has to be abstract to be reusable (so no HAL functions are allowed). Ideally it could be written as a standalone library directly located inside the project itself, so that if needed other can copy-paste it.

## Implementation Steps

1. Read the data sheet and understand how the led works
2. Define how the microcontroller should handle them
3. Document everything

## Dependencies

None

## Additional Notes

None

### LINKED ISSUE #6: Implement LEDs module
# Description

On the steering wheel there are 9 LEDs, divided in 3 groups:
- top left: 2 LEDs
- top right: 2 LEDs
- middle: 5 LEDs

Based on the situation these LEDs will have a different color, brightness and pattern.

The module to implement handles all this cases while being hardware-agnostic, so that it can be bound to a driver later.

## Implementation Steps

1. write down all possible states of the led based on every possible situation
2. implement the actual module, keeping in mind coding conventions
3. create tests for the module

## Dependencies

None

## Additional Notes

None

Bridiro:
# Purpose
Add an abstract led module that can perform actions needed by the steering wheel interface.
At the same time, a WS2812B driver was needed, as it is the model used by the steering wheel itself.

# Overview
Added 2 main modules:
- led -> abstract, holds led states
- ws2812b -> driver that provides a translation layer for led module to comunicate to hardware

# Guidance
Check added files for compliance with our rules, then try to run tests.

This PR may close both #3 and #6


CM4/Core/Inc/steering/drivers/ws2812b.h
```diff
@@ -0,0 +1,27 @@
...
+
+#ifndef WS2812B_H
+#define WS2812B_H
+
+#include <stdint.h>
+#include <stddef.h>
+
+#define WS2812B_DUTY_0 35
+#define WS2812B_DUTY_1 70
+#define WS2812B_RESET_SLOTS 50
```
Tonidotpy: You can add the `U` prefix if you like, may help when comparing with unsigned identifiers.
```suggestion
#define WS2812B_DUTY_0 (35)
#define WS2812B_DUTY_1 (70)
#define WS2812B_RESET_SLOTS (50)
```

CM4/Core/Inc/steering/drivers/ws2812b-api.h
```diff
@@ -0,0 +1,29 @@
...
+
+#include "ws2812b.h"
+
+/*!
+ * \brief Encodes RGB color data into a PWM buffer suitable for WS2812B LEDs.
+ *
+ * This function takes an array of RGB color values and converts them into a format that can be transmitted to WS2812B LEDs. The output buffer will contain the appropriate PWM values for each bit of the color data, as well as the necessary reset slots at the end.
+ *
+ * \param brightness Brightness level for the LEDs (0-255), where 0 is off and 255 is full brightness (scaling is internal)
+ * \param input_bgr Pointer to an array of GRB color values (3 bytes per LED, in GRB order).
```
Tonidotpy: Why in the brief you are talking about RGB format, the variable contains BGR in the name but its description tells that it uses GRB?? It is kinda confusing to be honest.
Also for color formats you should use an union/struct that automatically handles multiple channels; this can be made common to both cores.

Here is an example, it also handles endiannes but it should be tested to ensure that the channel order is preserved for any architecture:
```c
union Color {
    uint32_t rgba;
    struct {
#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
        uint8_t a, b, g, r;
#else
        uint8_t r, g, b, a;
#endif
    } channels;
};
```

CM4/Core/Src/steering/drivers/ws2812b-api.c
```diff
@@ -0,0 +1,69 @@
...
+};
+// clang-format on
+
+enum WS2812BReturnCode ws2812b_encode(
+    uint8_t brightness,
+    const uint8_t *input,
+    uint16_t *output,
+    size_t num_leds) {
+
+    if (!input || !output)
```
Tonidotpy: Always compare against `NULL` as written in the conventions...

```suggestion
    if (input == NULL || output == NULL)
```

CM4/Core/Src/steering/drivers/ws2812b-api.c
```diff
@@ -0,0 +1,69 @@
...
+
+#include "ws2812b-api.h"
+#include "eagletrt.h"
+
+// clang-format off
+
+/*!
+ * \brief Gamma correction table for 8-bit color values.
+ */
+EAGLETRT_STATIC const uint8_t GAMMA[256] = {
```
Tonidotpy: I suggest a more descriptive name also following the convention since it seems more like a macro written like this.

```suggestion
EAGLETRT_STATIC const uint8_t GammaCorrectionTable[256] = {
```

CM4/Core/Src/steering/drivers/ws2812b-api.c
```diff
@@ -0,0 +1,69 @@
...
+    size_t in_idx = 0;
+
+    for (size_t led = 0; led < num_leds; led++) {
+        /* 3 bytes per LED: G, R, B */
+        for (int byte = 0; byte < 3; byte++) {
+            uint8_t value = ((uint16_t)GAMMA[input[in_idx++]] * brightness) / 255;
+            /* MSB first */
+            for (int bit = 7; bit >= 0; bit--) {
+                output[out_idx++] =
+                    (value & (1 << bit)) ? WS2812B_DUTY_1 : WS2812B_DUTY_0;
```
Tonidotpy: Bit manipulation can be made easier to read (and safer since it is already tested) using the eagletrt library macros. Take a look at them, there is a specific macro perfectly suited for this type of operation.

Tonidotpy: I'll avoid incrementing inside the array access operator, it is difficult to read and may cause problems if not handled correctly:
```suggestion
                output[out_idx] =
                    (value & (1 << bit)) ? WS2812B_DUTY_1 : WS2812B_DUTY_0;
                ++out_idx;
```

CM4/Core/Src/steering/drivers/ws2812b-api.c
```diff
@@ -0,0 +1,69 @@
...
+    if (!input || !output)
+        return WS2812B_RC_NULL_POINTER;
+
+    size_t out_idx = 0;
+    size_t in_idx = 0;
+
+    for (size_t led = 0; led < num_leds; led++) {
+        /* 3 bytes per LED: G, R, B */
+        for (int byte = 0; byte < 3; byte++) {
+            uint8_t value = ((uint16_t)GAMMA[input[in_idx++]] * brightness) / 255;
```
Tonidotpy: I'll avoid incrementing inside the array access operator, it is difficult to read and may cause problems if not handled correctly:
```suggestion
            uint8_t value = ((uint16_t)GAMMA[input[in_idx]] * brightness) / 255;
            ++in_idx;
```

CM4/Core/Src/steering/drivers/ws2812b-api.c
```diff
@@ -0,0 +1,69 @@
...
+            for (int bit = 7; bit >= 0; bit--) {
+                output[out_idx++] =
+                    (value & (1 << bit)) ? WS2812B_DUTY_1 : WS2812B_DUTY_0;
+            }
+        }
+    }
+
+    /* reset (low for >50µs) */
+    for (size_t i = 0; i < WS2812B_RESET_SLOTS; i++) {
+        output[out_idx++] = 0;
```
Tonidotpy: Same here, avoid incrementing inside array access operators:
```suggestion
        output[out_idx] = 0;
        ++out_idx;
```

CM4/Core/Src/steering/drivers/ws2812b-api.c
Tonidotpy, line 1: Watch out for buffer overflows, you have to be sure that the indices values do not exceeds the maximum length of the array.

CM4/Core/Inc/steering/drivers/ws2812b-api.h
```diff
@@ -0,0 +1,29 @@
...
+ * \param output_pwm Pointer to a buffer where the encoded PWM values will be stored. The buffer must be large enough to hold the encoded data for all LEDs plus the reset slots.
+ * \param num_leds The number of LEDs in the input array.
+ *
+ * \retval WS2812B_RC_OK Encoding successful.
+ * \retval WS2812B_RC_NULL_POINTER A null pointer was passed for either input_rgb or output_pwm.
+ */
+enum WS2812BReturnCode ws2812b_encode(uint8_t brightness, const uint8_t *input_grb, uint16_t *output_pwm, size_t num_leds);
```
Tonidotpy: `input` and `output` identifier names are too generic, everything is either an input or an output.
The input should be called color or something equivalent since that's what it is. The output is not really clear, what is it of a PWM signal? The frequency? The duty cycle? The amplitude? A status?

CM4/Core/Src/steering/fsm.c
```diff
@@ -121,7 +126,11 @@ fsm_state_t fsm_do_idle(fsm_state_data_t *data) {
 
     /*** USER CODE BEGIN DO_IDLE ***/
 
-    if (inputs_api_poll_for_long_press(&input_handler, HAL_GetTick()) != INPUTS_RC_OK) {
+    if (inputs_api_poll_for_long_press(HAL_GetTick()) != INPUTS_RC_OK) {
```
Tonidotpy: No `HAL` functions inside hardware independent code such as the FSM

CM4/Core/Src/steering/fsm.c
```diff
@@ -21,7 +21,8 @@ Functions and types have been generated with prefix "fsm_"
 #include "eagletrt-api.h"
 #include "inputs-api.h"
 #include "ipc-api.h"
-#include "gpio.h"
+#include "leds-api.h"
+#include "tim.h"
 #include "main.h"
```
Tonidotpy: Avoid inclusion of `main.h`, causes a lot of troubles since it recursively includes almost everything (especially the drivers).

CM4/Core/Inc/steering/leds.h
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+#ifndef LEDS_H
+#define LEDS_H
+
+#include "ws2812b.h"
+#include <stdint.h>
+#include <stddef.h>
+
+#define LEDS_COUNT 9
```
Tonidotpy:
```suggestion
#define LEDS_COUNT (9U)
```

CM4/Core/Inc/steering/leds.h
```diff
@@ -0,0 +1,78 @@
...
+ * \retval LEDS_RC_TRANSMISSION_ERROR if an error occurred during data transmission.
+ */
+typedef enum LedsReturnCode (*leds_transmit_callback)(const uint16_t *buffer, size_t size);
+
+/*!
+ * \brief Handler structure for managing the LED system.
+ */
+struct LedsHandler {
+    struct LedColor leds[LEDS_COUNT];          /*!< Array of LedColor structures representing the colors of each LED. */
+    uint16_t pwm_buffer[LEDS_PWM_BUFFER_SIZE]; /*!< Buffer for storing the PWM data to be transmitted to the LEDs. */
```
Tonidotpy: PWM transmission is hardware dependent, the led module should handle input and output data independently from the peripheral used for the LEDs.
PWM signal transmission should be handled by the hardware dependent code given the necessary data by the LED module.

For example if the duty cycle changes the brightness of the LEDs the LED module should handle the brightness value, it is the HW dependent code job to map the brightness value to the right PWM signal values and timings.

For this type of LEDs I think the parameters are:
- their **colors**
- their **brightness**
- their **status**? (either on or off but may be controlled also by the brightness)

You can imagine for example as if the same model of LED strip uses SPI (with a specific command completely different from the PWM) instead of PWM to communicate. **What portion of code should you change right now inside the LED module?**

CM4/Core/Inc/steering/leds-api.h
```diff
@@ -0,0 +1,112 @@
...
+/*!
+ * \brief Transmits the current LED color data to the LED strip, updating the displayed colors.
+ *
+ * \param brightness Brightness level for the LEDs (0-255), where 0 is off and 255 is full brightness (scaling is internal)
+ *
+ * \retval LEDS_RC_OK Data transmitted successfully.
+ * \retval LEDS_RC_NULL_POINTER The transmit callback function is not set.
+ * \retval LEDS_RC_TRANSMISSION_ERROR An error occurred during data transmission.
+ */
+enum LedsReturnCode leds_api_show(uint8_t brightness);
```
Tonidotpy: It is a bit strange to set the brightness when you transmit the signal but the color can be set without transmission.
Also I don't think `show` is a good name for what this function is actually doing.

CM4/Core/Inc/steering/leds-api.h
```diff
@@ -0,0 +1,112 @@
...
+/*!
+ * \brief Sets the color of a specific LED in the strip.
+ *
+ * \param index Index of the LED to be set (0-based).
+ * \param color LedColor structure representing the desired color for the LED.
+ *
+ * \retval LEDS_RC_OK Color set successfully.
+ * \retval LEDS_RC_INVALID_LED The specified LED index is out of range.
+ */
+enum LedsReturnCode leds_api_set_led(enum LedsIndex index, struct LedColor color);
```
Tonidotpy: `set_led_color` may be better as a name or you can also provide the brightness instead of only the color.

CM4/Core/Inc/steering/leds-api.h
```diff
@@ -0,0 +1,112 @@
...
+
+/*!
+ * \brief Fills the entire LED strip with a specific color.
+ *
+ * \param color LedColor structure representing the desired color for all LEDs.
+ */
+void leds_api_fill(struct LedColor color);
```
Tonidotpy: Not really sure what am I filling, may be better to call it `set_led_color_all`. Brightness may be also provided here.

CM4/Core/Inc/steering/leds-api.h
```diff
@@ -0,0 +1,112 @@
...
+enum LedsReturnCode leds_api_show(uint8_t brightness);
+
+/*!
+ * \brief Sets the LEDs to indicate the PTT (Push-To-Talk) status.
+ *
+ * This function can be used to visually indicate when the PTT is active.
+ *
+ * \details The middle 5 LEDs (indices 0-4) will be set to blue
+ */
+void leds_api_ptt();
```
Tonidotpy: Name is not descriptive enough of the functionality

CM4/Core/Inc/steering/leds-api.h
```diff
@@ -0,0 +1,112 @@
...
+
+/*!
+ * \brief Sets the LEDs to indicate the current lap status.
+ *
+ * This function can be used to visually indicate the current lap number.
+ *
+ * \details The top left and right LEDs (indices 5-8) will be set to off (TODO: ask if green is better)
+ */
+void leds_api_target_lap(void);
```
Tonidotpy: Name is not descriptive enough of the functionality

CM4/Core/Inc/steering/leds-api.h
```diff
@@ -0,0 +1,112 @@
...
+
+/*!
+ * \brief Sets the LEDs to indicate the current lap status.
+ *
+ * This function can be used to visually indicate the lap delta is improving.
+ *
+ * \details The top left and right LEDs (indices 5-8) will be set to green (TODO: ask if green or purple)
+ */
+void leds_api_fast_lap(void);
```
Tonidotpy: Name is not descriptive enough of the functionality

CM4/Core/Inc/steering/leds-api.h
```diff
@@ -0,0 +1,112 @@
...
+
+/*!
+ * \brief Sets the LEDs to indicate the current lap status.
+ *
+ * This function can be used to visually indicate the lap delta is worsening.
+ *
+ * \details The top left and right LEDs (indices 5-8) will be set to yellow
+ */
+void leds_api_slow_lap(void);
```
Tonidotpy: Name is not descriptive enough of the functionality

CM4/Core/Inc/steering/leds-api.h
```diff
@@ -0,0 +1,112 @@
...
+
+/*!
+ * \brief Sets the LEDs to indicate an error state.
+ *
+ * This function can be used to visually indicate that an error has occurred.
+ *
+ * \details The middle 5 LEDs (indices 0-4) will be set to red
+ */
+void leds_api_error(void);
```
Tonidotpy: Name is not descriptive enough of the functionality

CM4/Core/Inc/steering/leds-api.h
```diff
@@ -0,0 +1,112 @@
...
+
+/*!
+ * \brief Sets the LEDs to indicate an OK state.
+ *
+ * This function can be used to visually indicate that everything is functioning correctly.
+ *
+ * \details The middle 5 LEDs (indices 0-4) will be set to off (TODO: ask if green is better)
+ */
+void leds_api_ok(void);
```
Tonidotpy: Name is not descriptive enough of the functionality

CM4/Core/Src/steering/leds-api.c
```diff
@@ -0,0 +1,114 @@
...
+EAGLETRT_STATIC struct LedsHandler leds_handler;
+
+enum LedsReturnCode leds_api_init(leds_transmit_callback transmit) {
+    if (transmit == NULL) {
+        return LEDS_RC_NULL_POINTER;
+    }
+
+    leds_handler.transmit = transmit;
+    memset(leds_handler.leds, 0, sizeof(leds_handler.leds));
+    memset(leds_handler.pwm_buffer, 0, sizeof(leds_handler.pwm_buffer));
```
Tonidotpy: You can `memset` the whole structure to 0 before setting any value.
It's easier, faster and works even if you modify the structure definition.

CM4/Core/Src/steering/leds-api.c
```diff
@@ -0,0 +1,114 @@
...
+}
+
+void leds_api_fill(struct LedColor color) {
+    for (size_t i = 0; i < LEDS_COUNT; i++) {
+        leds_handler.leds[i] = color;
+    }
+}
+
+void leds_api_clear(void) {
+    struct LedColor off = { 0, 0, 0 };
```
Tonidotpy, line 43: Why setting the color to black when you can set the brightness to 0?

CM4/Core/Src/steering/leds-api.c
```diff
@@ -0,0 +1,114 @@
...
+    if (transmit_rc != LEDS_RC_OK) {
+        return LEDS_RC_TRANSMISSION_ERROR;
+    }
+
+    return LEDS_RC_OK;
+}
+
+void leds_api_ptt(void) {
+    struct LedColor blue = { .g = 0, .r = 0, .b = 255 };
+    for (size_t i = 0; i < 5; i++) {
```
Tonidotpy: Since `5` is an arbitrary number you should create a define for that, to give it a meaningful name

test/cm4/test_ws2812b/test_ws2812b.c
```diff
@@ -0,0 +1,184 @@
...
+    }
+}
+
+/*!
+ * \defgroup ws2812b_encode ws2812b_encode tests
+ * \{
+ */
+
+void test_ws2812b_encode_null_input(void) {
+    uint16_t output[ONE_LED_OUTPUT_SIZE];
```
Tonidotpy: Uninitialized variables may cause unexpected behaviours.
```suggestion
    uint16_t output[ONE_LED_OUTPUT_SIZE] = { 0 };
```

test/cm4/test_ws2812b/test_ws2812b.c
```diff
@@ -0,0 +1,184 @@
...
+ *
+ * \brief Unit tests for the WS2812B encoding functionality.
+ */
+
+#include "unity.h"
+#include "ws2812b-api.h"
+#include <string.h>
+#include <stdio.h>
+
+#define ONE_LED_INPUT_SIZE 3
```
Tonidotpy:
```suggestion
#define ONE_LED_INPUT_SIZE (3U)
```

CM4/Core/Inc/steering/drivers/ws2812b.h
Tonidotpy, line 1: Are you sure that no handler structure is needed for this driver?

test/cm4/test_ws2812b/test_ws2812b.c
Tonidotpy, line 1: If an handler structure is needed remember to call and test the `init` function for the driver

test/cm4/test_ws2812b/test_ws2812b.c
```diff
@@ -0,0 +1,184 @@
...
+#define ONE_LED_OUTPUT_SIZE (ONE_LED_INPUT_SIZE * 8 + WS2812B_RESET_SLOTS)
+
+extern const uint8_t GAMMA[256];
+
+static void encode_byte(uint8_t value, uint16_t *out, uint8_t brightness) {
+    uint8_t scaled_value = ((uint16_t)GAMMA[value] * brightness) / 255;
+    for (int i = 7; i >= 0; i--) {
+        out[7 - i] = (scaled_value & (1 << i)) ? WS2812B_DUTY_1 : WS2812B_DUTY_0;
```
Tonidotpy: Eagletrt library contains macros for bit manipulation, check them.

test/cm4/test_ws2812b/test_ws2812b.c
```diff
@@ -0,0 +1,184 @@
...
+
+void test_ws2812b_encode_null_input(void) {
+    uint16_t output[ONE_LED_OUTPUT_SIZE];
+    uint8_t brightness = 255;
+    enum WS2812BReturnCode rc = ws2812b_encode(brightness, NULL, output, 1);
+    TEST_ASSERT_EQUAL(WS2812B_RC_NULL_POINTER, rc);
+}
+
+void test_ws2812b_encode_null_output(void) {
+    uint8_t input[ONE_LED_INPUT_SIZE] = { 0xFF, 0x00, 0x00 }; // Red
```
Tonidotpy: Why is the color described by an array of `uint8_t` rather than a structure?

test/cm4/test_ws2812b/test_ws2812b.c
```diff
@@ -0,0 +1,184 @@
...
+    uint8_t input[3] = {
+        0xAA,
+        0x55,
+        0xFF
+    };
+
+    uint8_t brightness = 255;
+
+    uint16_t output[ONE_LED_OUTPUT_SIZE];
+    memset(output, 0xFFFF, sizeof(output));
```
Tonidotpy: **!!! DANGEROUS CODE HERE !!!**

If you try the following code you'll understand why:
```
uint16_t a[3];
    memset(a, 0x1234, sizeof(a));
    for (int i = 0; i < 3; ++i)
        printf("%x\n", a[i]);
```
In the example you expect to find the following output:
```
1234
1234
1234
```
but what you get is:
```
3434
3434
3434
```
That is not the intended way of using `memset`, coincidently with the value of `0xffff` works because both bytes have the same value.

test/cm4/test_ws2812b/test_ws2812b.c
```diff
@@ -0,0 +1,184 @@
...
+
+    enum WS2812BReturnCode rc = ws2812b_encode(brightness, input, output, 1);
+    TEST_ASSERT_EQUAL_MESSAGE(
+        WS2812B_RC_OK,
+        rc,
+        "ws2812b_encode should return WS2812B_RC_OK for valid input and output pointers");
```
Tonidotpy: Assertion should be at the end of the test.
Remember the order:
1. Arrange
2. Act
3. Assert
Never mix the three, it's confusing.

test/cm4/test_ws2812b/test_ws2812b.c
```diff
@@ -0,0 +1,184 @@
...
+    ptr += 8;
+    encode_byte(0x55, ptr, brightness);
+    ptr += 8;
+    encode_byte(0xFF, ptr, brightness);
+    ptr += 8;
+
+    for (size_t i = 0; i < WS2812B_RESET_SLOTS; i++) {
+        *ptr++ = 0;
+    }
+
```
Tonidotpy: Don't really like this part, not sure what to do here.
We can discuss about it.

test/cm4/test_ws2812b/test_ws2812b.c
```diff
@@ -0,0 +1,184 @@
...
+    TEST_ASSERT_EQUAL_UINT16_ARRAY_MESSAGE(expected, output, ONE_LED_OUTPUT_SIZE, "ws2812b_encode should produce the correct PWM pattern for a single LED input");
+}
+
+void test_ws2812b_encode_two_leds(void) {
```
Tonidotpy, line 89: Not really needed to test one or two leds.
What you are interested in is if the behaviour using valid data is correct or not. You can just try with multiple leds.

test/cm4/test_ws2812b/test_ws2812b.c
```diff
@@ -0,0 +1,184 @@
...
+
+void test_ws2812b_encode_msb_first(void) {
+    uint8_t input[3] = {
+        0x80, /* 10000000 => gamma corrected => 01001001 */
+        0x01, /* 00000001 => gamma corrected => 00000000 */
+        0x00
+    };
+
+    printf("GAMMA[0x80] = %u\n", GAMMA[0x80]);
+    printf("GAMMA[0x01] = %u\n", GAMMA[0x01]);
```
Tonidotpy: ???

test/cm4/test_leds/test_leds.c
```diff
@@ -0,0 +1,252 @@
...
+extern struct LedsHandler leds_handler;
+
+DEFINE_FFF_GLOBALS;
+
+FAKE_VALUE_FUNC(enum LedsReturnCode, fake_transmit, const uint16_t *, size_t);
+
+void setUp(void) {
+    RESET_FAKE(fake_transmit);
+    FFF_RESET_HISTORY();
+    memset(&leds_handler, 0, sizeof(leds_handler));
```
Tonidotpy: You should call `leds_api_init` here.

test/cm4/test_leds/test_leds.c
```diff
@@ -0,0 +1,252 @@
...
+    TEST_ASSERT_EQUAL_MESSAGE(LEDS_RC_OK, rc, "leds_api_init should return LEDS_RC_OK on successful initialization");
+    TEST_ASSERT_EQUAL_PTR_MESSAGE(fake_transmit, leds_handler.transmit, "leds_handler.transmit should be set to the provided transmit function");
+}
+
+void test_leds_api_init_null_pointer(void) {
+    enum LedsReturnCode rc = leds_api_init(NULL);
+    TEST_ASSERT_EQUAL(LEDS_RC_NULL_POINTER, rc);
+}
+
+void test_leds_api_init_clears_leds(void) {
```
Tonidotpy: The content of the handler should be checked inside the first test with valid inputs.

Tonidotpy (changes requested):
Avoid pushing changes of other modules in a PR for the LED module, it's harder for the reviewer to do his job correctly.
The way that colors and brightness are handled is a bit strange in my opinion. For example if I only want to set the brightness I need to also send the data to the pyshical LEDs and the value is not actually handled by the LED module but by the driver. This differs if I want to do the same with the LED colors.

CM4/Core/Inc/steering/drivers/ws2812b-api.h
Bridiro: I actually wrote the structure exactly like that because, without any computational effort, I can cast it to (uint8_t *) and pass it directly to the encoder function (as it needs values like this: [G, R, B, G, R, B, ...])

CM4/Core/Src/steering/fsm.c
Bridiro: I found no other way of accessing the `__DMB` function. I have literally no other idea on how to do it

CM4/Core/Inc/steering/drivers/ws2812b-api.h
Tonidotpy: Where do you wrote it? Because that's not what I saw in the PR

CM4/Core/Src/steering/fsm.c
Bridiro: Same here as it was for the `__DMB` function. How do I access it? Do I have to write callbacks for the FSM as well?

Tonidotpy: In the HV BMS firmware I used callbacks, you can also pass user data as argument to the FSM state functions

CM4/Core/Inc/steering/drivers/ws2812b-api.h
Bridiro: it is at `CM4/Core/Src/steering/leds-api.c:53`. That is the best solution I found that did not require any computation at all (loops to re-arrange data or so)

test/cm4/test_ws2812b/test_ws2812b.c
Bridiro: that's because this driver accepts an array of `uint8_t` as in input, ordered GRB (yeah I made an error in the comment but that's not the point). The one handling the "easy" interface with the structure is the led module.

CM4/Core/Inc/steering/leds.h
```diff
@@ -0,0 +1,63 @@
...
+    LEDS_CENTER_1 = 1,    /*!< Center LED 1 (index 1). */
+    LEDS_CENTER_2 = 2,    /*!< Center LED 2 (index 2). */
+    LEDS_CENTER_3 = 3,    /*!< Center LED 3 (index 3). */
+    LEDS_CENTER_4 = 4,    /*!< Center LED 4 (index 4). */
+    LEDS_TOP_LEFT_1 = 5,  /*!< Top left LED 1 (index 5). */
+    LEDS_TOP_LEFT_0 = 6,  /*!< Top left LED 0 (index 6). */
+    LEDS_TOP_RIGHT_0 = 7, /*!< Top right LED 0 (index 7). */
+    LEDS_TOP_RIGHT_1 = 8, /*!< Top right LED 1 (index 8). */
+    LEDS_COUNT = 9        /*!< Total number of LEDs in the strip. */
+};
```
Tonidotpy, line 67: **Remember the prefix!!**
Enums value increments automatically, you can explicitely set the first value to 0 to mark the intent of the numbers to be from 0 to n.
```suggestion
enum LedsIndex {
    LEDS_INDEX_CENTER_0 = 0, /*!< Center LED 0. */
    LEDS_INDEX_CENTER_1,     /*!< Center LED 1. */
    LEDS_INDEX_CENTER_2,     /*!< Center LED 2. */
    LEDS_INDEX_CENTER_3,     /*!< Center LED 3. */
    LEDS_INDEX_CENTER_4,     /*!< Center LED 4. */
    LEDS_INDEX_TOP_LEFT_1,   /*!< Top left LED 1. */
    LEDS_INDEX_TOP_LEFT_0,   /*!< Top left LED 0. */
    LEDS_INDEX_TOP_RIGHT_0,  /*!< Top right LED 0. */
    LEDS_INDEX_TOP_RIGHT_1,  /*!< Top right LED 1. */
    LEDS_INDEX_COUNT         /*!< Total number of LEDs in the strip. */
};
```

CM4/Core/Inc/steering/drivers/ws2812b-api.h
```diff
@@ -0,0 +1,49 @@
...
+
+#ifndef WS2812B_API_H
+#define WS2812B_API_H
+
+#include "ws2812b.h"
+
+/*!
+ * \brief Macro to calculate the required buffer size for encoding a given number of WS2812B LEDs.
+ */
+#define WS2812B_API_BUFFER_SIZE(num_leds) ((num_leds) * 24 + WS2812B_RESET_SLOTS)
```
Tonidotpy: I usually write every identifier uppercase in a macro. Just a personal preference, not really necessary.

CM4/Core/Inc/steering/drivers/ws2812b.h
```diff
@@ -0,0 +1,50 @@
...
+#define WS2812B_H
+
+#include <stdint.h>
+#include <stddef.h>
+#include <stdbool.h>
+
+#define WS2812B_FREQUENCY_HZ (800000U) /*!< Frequency in Hz for WS2812B LED data transmission. */
+#define WS2812B_RESET_SLOTS (50U)      /*!< Number of time slots required to reset the WS2812B LEDs after data transmission. */
+#define WS2812B_DUTY_0_RATIO (35U)     /*!< Duty cycle ratio for representing binary 0 (35% of the time slot). */
+#define WS2812B_DUTY_1_RATIO (70U)     /*!< Duty cycle ratio for representing binary 1 (70% of the time slot). */
```
Tonidotpy: Since those are ratios they should be floating point number from 0 to 1, it's up to you to decide if you want to keep them as number from 0 to 100 and do the conversion later. Btw, percentages are also from 0 to 1 to be precise.

```suggestion
#define WS2812B_DUTY_0_RATIO (0.35f)     /*!< Duty cycle ratio for representing binary 0 (35% of the time slot). */
#define WS2812B_DUTY_1_RATIO (0.7f)     /*!< Duty cycle ratio for representing binary 1 (70% of the time slot). */
```

CM4/Core/Inc/steering/drivers/ws2812b-api.h
```diff
@@ -0,0 +1,49 @@
...
+ */
+enum WS2812BReturnCode ws2812b_api_init(struct WS2812BHandler *handler, ws2812b_get_tick_hz_callback get_tick_hz_callback);
+
+/*!
+ * \brief Encodes GRB color data into a PWM buffer suitable for WS2812B LEDs.
+ *
+ * This function takes an array of GRB color values and converts them into a format that can be transmitted to WS2812B LEDs. The output buffer will contain the appropriate PWM values for each bit of the color data, as well as the necessary reset slots at the end.
+ *
+ * \param handler Pointer to an initialized WS2812BHandler structure containing the timing parameters for encoding.
+ * \param brightness Brightness level for the LEDs (0-255), where 0 is off and 255 is full brightness (scaling is internal)
```
Tonidotpy: It may be better, at least logically, to have the brightness going from 0 to 1 (0% to 100%) instead of 0 to 255 since it is not a color channel.
It also simplify calculation since it completely avoids integer overflow (check line 70 in `ws2812b-api.c`)

CM4/Core/Inc/steering/drivers/ws2812b-api.h
```diff
@@ -0,0 +1,49 @@
...
+ * \param grb_color_buffer Pointer to an array of GRB color values (3 bytes per LED, in GRB order).
+ * \param pwm_duty_out Pointer to a buffer where the encoded PWM values will be stored. The buffer must be large enough to hold the encoded data for all LEDs plus the reset slots.
+ * \param num_leds The number of LEDs in the input array.
```
Tonidotpy: I'd prefer `led_count`, it make more sense to me. Just a suggestion, do as you please.

CM4/Core/Src/steering/drivers/ws2812b-api.c
```diff
@@ -0,0 +1,93 @@
...
+
+    if (handler == NULL || grb_color_buffer == NULL || pwm_duty_out == NULL)
+        return WS2812B_RC_NULL_POINTER;
+
+    size_t out_idx = 0;
+    size_t in_idx = 0;
+
+    for (size_t led = 0; led < num_leds; led++) {
+        // 3 bytes per LED: G, R, B
+        for (int byte = 0; byte < 3; byte++) {
```
Tonidotpy: They are usually called `channels`.
That's because a single channel may be represented with multiple bytes. E.g. classic RGB uses 4 bytes for each channel for a total of 12 bytes.

CM4/Core/Inc/steering/drivers/ws2812b-api.h
```diff
@@ -0,0 +1,49 @@
...
+
+/*!
+ * \brief Encodes GRB color data into a PWM buffer suitable for WS2812B LEDs.
+ *
+ * This function takes an array of GRB color values and converts them into a format that can be transmitted to WS2812B LEDs. The output buffer will contain the appropriate PWM values for each bit of the color data, as well as the necessary reset slots at the end.
+ *
+ * \param handler Pointer to an initialized WS2812BHandler structure containing the timing parameters for encoding.
+ * \param brightness Brightness level for the LEDs (0-255), where 0 is off and 255 is full brightness (scaling is internal)
+ * \param grb_color_buffer Pointer to an array of GRB color values (3 bytes per LED, in GRB order).
+ * \param pwm_duty_out Pointer to a buffer where the encoded PWM values will be stored. The buffer must be large enough to hold the encoded data for all LEDs plus the reset slots.
```
Tonidotpy: You should specify that the `WS2812B_API_BUFFER_SIZE` macro can be used to get the correct size for the buffer.

CM4/Core/Inc/steering/drivers/ws2812b-api.h
```diff
@@ -0,0 +1,49 @@
...
+ * \param num_leds The number of LEDs in the input array.
+ *
+ * \retval WS2812B_RC_OK Encoding successful.
+ * \retval WS2812B_RC_NULL_POINTER A null pointer was passed for either the input color buffer or the output PWM buffer.
+ */
+enum WS2812BReturnCode ws2812b_encode(struct WS2812BHandler *handler, uint8_t brightness, const uint8_t *grb_color_buffer, uint16_t *pwm_duty_out, size_t num_leds);
```
Tonidotpy: I'd rather return the number of encoded bytes (0 on error) instead of an enum error.
The returned value can be really useful for the user and it can also be compared to the buffer size to check for encoding errors.

CM4/Core/Src/steering/drivers/ws2812b-api.c
```diff
@@ -0,0 +1,93 @@
...
+        return WS2812B_RC_NULL_POINTER;
+    }
+
+    memset(handler, 0, sizeof(*handler));
+    uint32_t arr = (get_tick_hz_callback() / WS2812B_FREQUENCY_HZ) - 1;
+    handler->duty_0 = (uint16_t)((arr + 1) * WS2812B_DUTY_0_RATIO) / 100;
+    handler->duty_1 = (uint16_t)((arr + 1) * WS2812B_DUTY_1_RATIO) / 100;
+
+    return WS2812B_RC_OK;
+}
```
Tonidotpy: There is no need to use a function callback to get the frequency since you can just give as parameter the frequency itself 😂 
```suggestion
enum WS2812BReturnCode ws2812b_api_init(struct WS2812BHandler *handler, uint32_t frequency) {
    if (handler == NULL) {
        return WS2812B_RC_NULL_POINTER;
    }

    memset(handler, 0, sizeof(*handler));
    float arr = frequency / (float)WS2812B_FREQUENCY_HZ;
    handler->duty_0 = arr * WS2812B_DUTY_0_RATIO;
    handler->duty_1 = arr * WS2812B_DUTY_1_RATIO;

    return WS2812B_RC_OK;
}
```

CM4/Core/Inc/steering/drivers/ws2812b.h
```diff
@@ -0,0 +1,50 @@
...
+};
+
+/*!
+ * \brief Handler structure for WS2812B LED control.
+ *
+ * This structure contains the duty cycle values for representing binary 0 and 1, as well as the number of reset slots required after transmitting data to the LEDs.
+ */
+struct WS2812BHandler {
+    uint16_t duty_0; /*!< Duty cycle value representing binary 0. */
+    uint16_t duty_1; /*!< Duty cycle value representing binary 1. */
```
Tonidotpy: Since they are duty cycles they should be float values from 0 to 1 since they are percentages.

test/cm4/test_ws2812b/test_ws2812b.c
```diff
@@ -0,0 +1,221 @@
...
+#include "eagletrt-api.h"
+#include <stdio.h>
+
+#define TEST_FREQUENCY_HZ (80000000U) // 80 MHz timer clock for testing
+#define ONE_LED_INPUT_SIZE (3U)
+#define ONE_LED_OUTPUT_SIZE (ONE_LED_INPUT_SIZE * 8 + WS2812B_RESET_SLOTS)
+
+extern const uint8_t WS2812BGammaCorrectionTable[256];
+struct WS2812BHandler ws2812b_handler;
+enum WS2812BReturnCode init_rc;
```
Tonidotpy: Avoid using global variable such as the return code or similar it may cause unwanted behaviours inside during test execution. A single handler can be used globally since it is used in all tests and correctly cleared at the start of each test.

test/cm4/test_ws2812b/test_ws2812b.c
```diff
@@ -0,0 +1,221 @@
...
+
+DEFINE_FFF_GLOBALS;
+
+FAKE_VALUE_FUNC(uint32_t, fake_get_tick_hz);
```
Tonidotpy: Not needed if you give directly the frequency as parameter to the init function.

test/cm4/test_ws2812b/test_ws2812b.c
```diff
@@ -0,0 +1,221 @@
...
+ * \defgroup ws2812b_api_init ws2812b_api_init tests
+ * \{
+ */
+
+void test_ws2812b_api_init_success(void) {
+    TEST_ASSERT_EQUAL_MESSAGE(1, fake_get_tick_hz_fake.call_count, "ws2812b_api_init should call get_tick_hz exactly once");
+    TEST_ASSERT_EQUAL_MESSAGE(calculate_expected_duty(WS2812B_DUTY_0_RATIO), ws2812b_handler.duty_0, "handler.duty_0 should be calculated based on the tick frequency");
+    TEST_ASSERT_EQUAL_MESSAGE(calculate_expected_duty(WS2812B_DUTY_1_RATIO), ws2812b_handler.duty_1, "handler.duty_1 should be calculated based on the tick frequency");
+    TEST_ASSERT_EQUAL_MESSAGE(WS2812B_RC_OK, init_rc, "ws2812b_api_init should return WS2812B_RC_OK on successful initialization");
+}
```
Tonidotpy: Remember to always Arrange, Act and Assert in each test.
If you are testing the `init` function you should call it inside the test, as it is now the result depends on what happens outside of the test itself which is an unwanted behaviour of Unit Testing.

test/cm4/test_ws2812b/test_ws2812b.c
```diff
@@ -0,0 +1,221 @@
...
+    enum WS2812BReturnCode rc = ws2812b_encode(NULL, brightness, input, output, 1);
+    TEST_ASSERT_EQUAL(WS2812B_RC_NULL_POINTER, rc);
+}
+
+void test_ws2812b_encode_null_input(void) {
+    uint16_t output[ONE_LED_OUTPUT_SIZE] = { 0 };
+    uint8_t brightness = 255;
+    enum WS2812BReturnCode rc = ws2812b_encode(&ws2812b_handler, brightness, NULL, output, 1);
+    TEST_ASSERT_EQUAL(WS2812B_RC_NULL_POINTER, rc);
+}
```
Tonidotpy: It may be good to also check in both tests if the output array does not change.

test/cm4/test_ws2812b/test_ws2812b.c
```diff
@@ -0,0 +1,221 @@
...
+    ptr += 8;
+    encode_byte(0x00, ptr, brightness);
+    ptr += 8;
+
+    encode_byte(0xFF, ptr, brightness);
+    ptr += 8;
+    encode_byte(0x00, ptr, brightness);
+    ptr += 8;
+    encode_byte(0xFF, ptr, brightness);
+    ptr += 8;
```
Tonidotpy: Remember that for loops do exists...

test/cm4/test_ws2812b/test_ws2812b.c
```diff
@@ -0,0 +1,221 @@
...
+    uint8_t scaled_value = ((uint16_t)WS2812BGammaCorrectionTable[value] * brightness) / 255;
+    for (int i = 7; i >= 0; i--) {
+        out[7 - i] = EAGLETRT_API_BIT_GET(scaled_value, i) ? ws2812b_handler.duty_1 : ws2812b_handler.duty_0;
+    }
+}
+
+EAGLETRT_STATIC uint16_t calculate_expected_duty(uint32_t duty_ratio) {
+    uint32_t arr = (TEST_FREQUENCY_HZ / WS2812B_FREQUENCY_HZ) - 1;
+    return (uint16_t)((arr + 1) * duty_ratio) / 100;
+}
```
Tonidotpy: Don't really like additional logic inside tests, avoid whenever possible.
Also avoid mixing `static` and `EAGLETRT_STATIC`, **be consistent** on how you write code.
> **The worst possible code is the one that does not follow its own rules**

test/cm4/test_ws2812b/test_ws2812b.c
```diff
@@ -0,0 +1,221 @@
...
+        rc,
+        "ws2812b_encode should return WS2812B_RC_OK for valid input and output pointers");
+
+    TEST_ASSERT_EQUAL_UINT16_ARRAY_MESSAGE(expected, output, out_size, "ws2812b_encode should produce the correct PWM pattern for multiple LED input");
+}
+
+void test_ws2812b_encode_msb_first(void) {
+    uint8_t input[3] = {
+        0x80, /* 10000000 => gamma corrected => 01001001 */
+        0x01, /* 00000001 => gamma corrected => 00000000 */
```
Tonidotpy: You can also write them as `0b10000000` and `0b00000001` respectively. instead of using comments.

test/cm4/test_ws2812b/test_ws2812b.c
```diff
@@ -0,0 +1,221 @@
...
+        0x00
+    };
+
+    uint8_t brightness = 255;
+
+    uint16_t output[ONE_LED_OUTPUT_SIZE];
```
Tonidotpy: Never leave uninitialized variables:
```suggestion
    uint16_t output[ONE_LED_OUTPUT_SIZE] = { 0 };
```

test/cm4/test_ws2812b/test_ws2812b.c
```diff
@@ -0,0 +1,221 @@
...
+
+    ws2812b_encode(&ws2812b_handler, brightness, input, output, 1);
```
Tonidotpy: Missing check for the return code

test/cm4/test_ws2812b/test_ws2812b.c
```diff
@@ -0,0 +1,221 @@
...
+
+    TEST_ASSERT_EQUAL_MESSAGE(ws2812b_handler.duty_0, output[0], "First bit of the first byte should be encoded as handler.duty_0");
+    TEST_ASSERT_EQUAL_MESSAGE(ws2812b_handler.duty_1, output[1], "Second bit of the first byte should be encoded as handler.duty_1");
+    TEST_ASSERT_EQUAL_MESSAGE(ws2812b_handler.duty_0, output[15], "Last bit of the second byte should be encoded as handler.duty_0");
+}
+
+void test_ws2812b_encode_reset_slots_are_zero(void) {
+    uint8_t input[3] = { 0x00, 0x00, 0x00 };
+    uint8_t brightness = 255;
+    uint16_t output[ONE_LED_OUTPUT_SIZE];
```
Tonidotpy:
```suggestion
    uint16_t output[ONE_LED_OUTPUT_SIZE] = { 0 };
```

test/cm4/test_ws2812b/test_ws2812b.c
```diff
@@ -0,0 +1,221 @@
...
+    ws2812b_encode(&ws2812b_handler, brightness, input, output, 1);
+    size_t start = 24;
+
+    for (size_t i = 0; i < WS2812B_RESET_SLOTS; i++) {
+        char msg[50];
+        snprintf(msg, sizeof(msg), "Reset slot %zu should be 0", i);
+        TEST_ASSERT_EQUAL_UINT16_MESSAGE(0, output[start + i], msg);
+    }
```
Tonidotpy: Avoid asserts inside loops, the `ARRAY` variant exists for this purpose.
Also avoid using non literal strings since they may cause unwantend bugs that are difficult to find. (i.e. don't use `sprintf`)

test/cm4/test_ws2812b/test_ws2812b.c
```diff
@@ -0,0 +1,221 @@
...
+}
+
+void test_ws2812b_encode_brightness_scaling(void) {
+    uint8_t input[3] = { 0xFF, 0xFF, 0xFF };
+    uint8_t brightness = 128; // 50% brightness
+    uint16_t output[ONE_LED_OUTPUT_SIZE];
```
Tonidotpy:
```suggestion
    uint16_t output[ONE_LED_OUTPUT_SIZE] = { 0 };
```

test/cm4/test_ws2812b/test_ws2812b.c
```diff
@@ -0,0 +1,221 @@
...
+    ws2812b_encode(&ws2812b_handler, brightness, input, output, 1);
+}
```
Tonidotpy, line 173: Where are the asserts???

CM4/Core/Inc/steering/leds.h
```diff
@@ -0,0 +1,91 @@
...
+/*!
+ * \brief Handler structure for managing the LED system.
+ */
+struct LedsHandler {
+    struct LedColor leds[LEDS_COUNT];                     /*!< Array of LedColor structures representing the colors of each LED. */
+    uint8_t brightness;                                   /*!< Brightness level for the LEDs (0-255), where 0 is off and 255 is full brightness. (default: 255) */
+    uint16_t buffer[WS2812B_API_BUFFER_SIZE(LEDS_COUNT)]; /*!< Buffer for storing the PWM data to be transmitted to the LEDs. */
+    struct WS2812BHandler ws2812b_handler;                /*!< Handler for managing the WS2812B LED strip. */
+    leds_transmit_callback transmit_callback;             /*!< Callback function for transmitting the LED data to the hardware. */
+    leds_get_busy_callback get_busy_callback;             /*!< Callback function for checking if the handler is currently busy transmitting data. */
```
Tonidotpy: Is it really needed?
You want to know if it's busy when you try to transmit and the transmit callback does tell you if it is busy or not so there is not really a point here on keeping this callback.

CM4/Core/Inc/steering/leds.h
```diff
@@ -0,0 +1,91 @@
...
+    LEDS_TOP_RIGHT_0 = 7, /*!< Top right LED 0 (index 7). */
+    LEDS_TOP_RIGHT_1 = 8, /*!< Top right LED 1 (index 8). */
+    LEDS_COUNT = 9        /*!< Total number of LEDs in the strip. */
+};
+
+/*!
+ * \brief Handler structure for managing the LED system.
+ */
+struct LedsHandler {
+    struct LedColor leds[LEDS_COUNT];                     /*!< Array of LedColor structures representing the colors of each LED. */
```
Tonidotpy: I'd prefer `colors` as identifier name since that's what it contains:
```suggestion
    struct LedColor colors[LEDS_COUNT];                     /*!< Array of LedColor structures representing the colors of each LED. */
```

CM4/Core/Src/steering/leds-api.c
```diff
@@ -0,0 +1,118 @@
...
+
+    enum WS2812BReturnCode encode_rc = ws2812b_encode(
+        &leds_handler.ws2812b_handler,
+        leds_handler.brightness,
+        (const uint8_t *)leds_handler.leds,
+        leds_handler.buffer,
+        LEDS_COUNT);
+
+    if (encode_rc != WS2812B_RC_OK) {
+        return LEDS_RC_NULL_POINTER;
```
Tonidotpy: Returning this error doesn't really make sense, may be better to use the transmission error.

test/cm4/test_leds/test_leds.c
```diff
@@ -0,0 +1,276 @@
...
+ * \defgroup leds_api_init Test cases for leds_api_init function
+ * \{
+ */
+
+void test_leds_api_init_success(void) {
+    TEST_ASSERT_EQUAL_MESSAGE(LEDS_RC_OK, init_rc, "leds_api_init should return LEDS_RC_OK on successful initialization");
+    TEST_ASSERT_EQUAL_PTR_MESSAGE(fake_transmit, leds_handler.transmit_callback, "leds_api_init should set the transmit callback correctly");
+    TEST_ASSERT_EQUAL_PTR_MESSAGE(fake_get_busy, leds_handler.get_busy_callback, "leds_api_init should set the get_busy callback correctly");
+    TEST_ASSERT_EQUAL_MESSAGE(255, leds_handler.brightness, "leds_api_init should set the default brightness to 255");
+}
```
Tonidotpy, line 36: You are not calling the init function here, remember the Arrange, Act, Assert method.

test/cm4/test_leds/test_leds.c
```diff
@@ -0,0 +1,276 @@
...
+
+extern struct LedsHandler leds_handler;
+
+DEFINE_FFF_GLOBALS;
+
+FAKE_VALUE_FUNC(enum LedsReturnCode, fake_transmit, const uint32_t *, uint16_t);
+FAKE_VALUE_FUNC(bool, fake_get_busy);
+FAKE_VALUE_FUNC(uint32_t, fake_get_tick_hz);
+
+enum LedsReturnCode init_rc;
```
Tonidotpy: Avoid global variables, especially in tests.

test/cm4/test_leds/test_leds.c
```diff
@@ -0,0 +1,276 @@
...
+    struct LedColor color = { .r = 255, .g = 0, .b = 0 };               // Red
+    enum LedsReturnCode rc = leds_api_set_led_color(LEDS_COUNT, color); // Out of range index
+    TEST_ASSERT_EQUAL(LEDS_RC_INVALID_LED, rc);
+}
+
+void test_leds_api_set_led_color_negative_index(void) {
+    struct LedColor color = { .r = 255, .g = 0, .b = 0 };       // Red
+    enum LedsReturnCode rc = leds_api_set_led_color(-1, color); // Negative index
+    TEST_ASSERT_EQUAL(LEDS_RC_INVALID_LED, rc);
+}
```
Tonidotpy, line 73: You should also check if the led colors remains unchanged.

test/cm4/test_leds/test_leds.c
```diff
@@ -0,0 +1,276 @@
...
+
+void test_leds_api_clear(void) {
+    memset(&leds_handler.leds, 0xff, sizeof(leds_handler.leds));
+
+    leds_api_clear();
+    for (size_t i = 0; i < LEDS_COUNT; i++) {
+        TEST_ASSERT_EQUAL_UINT8_MESSAGE(0, leds_handler.leds[i].r, "leds_handler.leds[i].r should be cleared to 0");
+        TEST_ASSERT_EQUAL_UINT8_MESSAGE(0, leds_handler.leds[i].g, "leds_handler.leds[i].g should be cleared to 0");
+        TEST_ASSERT_EQUAL_UINT8_MESSAGE(0, leds_handler.leds[i].b, "leds_handler.leds[i].b should be cleared to 0");
+    }
```
Tonidotpy: Avoid asserts inside loops, there are the `MEMORY` and `ARRAY` variants for that.

test/cm4/test_leds/test_leds.c
```diff
@@ -0,0 +1,276 @@
...
+}
+
+void test_leds_api_set_ok_pattern(void) {
+    leds_api_set_ok_pattern();
+    for (size_t i = 0; i < 5; i++) {
+        TEST_ASSERT_EQUAL_UINT8_MESSAGE(0, leds_handler.leds[i].r, "leds_handler.leds[i].r should be set to 0 for OK indication");
+        TEST_ASSERT_EQUAL_UINT8_MESSAGE(0, leds_handler.leds[i].g, "leds_handler.leds[i].g should be set to 255 for OK indication");
+        TEST_ASSERT_EQUAL_UINT8_MESSAGE(0, leds_handler.leds[i].b, "leds_handler.leds[i].b should be set to 0 for OK indication");
+    }
+}
```
Tonidotpy, line 207: Check Unity `MEMORY` and `ARRAY` variants to avoid using asserts inside for loop, please.

Tonidotpy (changes requested):
- Avoid overuse of function pointers, always try to find a way to **NOT** use them.
- Unit tests looks better than before but there are a couple of bad implementations that needs to be fixed.

Keep on going, I'm starting to like how those modules are being developed 👍🏼

CM4/Core/Inc/steering/drivers/ws2812b-api.h
Bridiro: I don't like the idea. Returning the number of bytes obscures the error in case of null pointer. An encoding error it's nearly impossible, while a null pointer is much more common and should be properly reported

CM4/Core/Inc/steering/drivers/ws2812b.h
Bridiro: They don't because I'm pre-processing them in the init function, so that I don't calculate them again every cycle

CM4/Core/Inc/steering/leds.h
Bridiro: yes it is needed as I also use it to check if I can encode or not. I can't modify the buffer if it's busy

CM4/Core/Inc/steering/drivers/ws2812b-api.h
Tonidotpy: It is common for drivers functions to return the number of encoded, decoded, serialized, etc... bytes.
Since it is nearly impossible for the function to fail the error can be easily detected by comparing the expected number of bytes with the return value. If they are different than there is an error.

CM4/Core/Inc/steering/drivers/ws2812b.h
Tonidotpy: What does that mean? You just need to change the type from `uint16_t` to `float` and slightly change the preprocessing inside the init function.

CM4/Core/Inc/steering/leds.h
Tonidotpy: You're right, I forgot about that. With the method I told you earlier this problem should be solved.

Bridiro: exactly, I just deleted it. This night/tomorrow morning I'll push the updated version

CM4/Core/Inc/steering/drivers/ws2812b.h
Bridiro: with the new changes I don't have this problem anymore, but before I did this so that I didn't have to process anything inside the encoding loop and I could just assign instead of having to calculate duty from ARR

CM4/Core/Inc/steering/drivers/ws2812b-api.h
Bridiro: That's true, but if I return the number of encoded bytes and pass a NULL pointer to the function, I won't be able to understand the source of the error, while as it is now I have a way of understanding what's gone wrong

CM4/Core/Inc/steering/drivers/ws2812b-api.h
```diff
@@ -0,0 +1,52 @@
...
+/*!
+ * \brief Macro to calculate the duty cycle value for a given frequency and ratio.
+ *
+ * This macro computes the duty cycle value based on the specified frequency and ratio, which is essential for generating the correct PWM signals for WS2812B LEDs.
+ *
+ * \param FREQ The frequency of the timer or PWM signal in Hz.
+ * \param RATIO The duty cycle ratio (e.g., WS2812B_DUTY_0_RATIO or WS2812B_DUTY_1_RATIO).
+ * \return The calculated duty cycle value as a 16-bit unsigned integer.
+ */
+#define WS2812B_API_CALCULATE_DUTY_VALUE(FREQ, RATIO) ((uint16_t)(((float)(FREQ) / WS2812B_FREQUENCY_HZ) * RATIO))
```
Tonidotpy, line 34: I'd usually avoid casting inside macros, it sometimes causes strange behaviours. Should not be a problem here, I hope at least.

test/cm4/test_leds/test_leds.c
```diff
@@ -0,0 +1,242 @@
...
+
+/*!
+ * \defgroup Steering wheel specific LED patterns
+ * \{
+ */
+
+void test_leds_api_set_ptt_pattern(void) {
+    leds_api_set_ptt_pattern();
+    struct LedColor expected_color = { .r = 0, .g = 0, .b = 255 };
+    struct LedColor expected_colors[5] = { [0 ... 4] = expected_color };
```
Tonidotpy, line 170: The `...` operator is a GNU extension, not really a problem but may not work if using different compilers.
Don't know if Clang supports it, but probably it does.

CM4/Core/Src/tim.c
```diff
@@ -22,6 +22,29 @@
 
 /* USER CODE BEGIN 0 */
 
+#include "eagletrt.h"
+#include "leds.h"
+
+EAGLETRT_STATIC volatile bool leds_busy = false;
+EAGLETRT_STATIC uint16_t duty0;
+EAGLETRT_STATIC uint16_t duty1;
+EAGLETRT_STATIC uint16_t duty_buffer[WS2812B_API_BUFFER_SIZE(LEDS_INDEX_COUNT)] = { 0 };
```
Tonidotpy: It may be better to create a struct here, does not need to be defined in the header file.

CM4/Core/Src/tim.c
```diff
@@ -500,4 +526,54 @@ void HAL_TIM_Base_MspDeInit(TIM_HandleTypeDef *tim_baseHandle) {
...
+    }
+
+    leds_busy = true;
+    return LEDS_RC_OK;
+}
+
+uint32_t tim_ws2812b_get_timer_hz() {
+    return prv_timer_get_tick_hz(&htim3);
+}
+
```
Tonidotpy: Not needed, should be removed

CM4/Core/Inc/tim.h
```diff
@@ -56,6 +57,38 @@ void HAL_TIM_MspPostInit(TIM_HandleTypeDef *htim);
...
+enum LedsReturnCode tim_leds_transmit(const enum WS2812BDutyCycle *buffer, uint16_t size);
+
+/*!
+ * \brief Gets the timer frequency in Hz for the WS2812B LED transmission.
+ *
+ * This function calculates the timer frequency based on the configuration of TIM3, which is used for transmitting data to the WS2812B LEDs. The frequency is determined by the timer's clock source and prescaler settings.
+ *
+ * \return The timer frequency in Hz.
+ */
+uint32_t tim_ws2812b_get_timer_hz();
```
Tonidotpy: Also remove this, not needed outside of the TIM module.

Tonidotpy (changes requested):
Couple of change but nothing big, should be approved on the next review.

CM4/Core/Inc/steering/drivers/ws2812b-api.h
Bridiro: it was giving me a warning before casting. Documentation tells the user what to pass to this macro so it should not be a problem

test/cm4/test_leds/test_leds.c
Bridiro: I confirm that this works with clang as well (I'm on macOS). Anyway as this is used via PlatformIO it should use the downloaded toolchain that is in fact gcc

Tonidotpy (approved):

PR #15, opened 2026-03-29
Task/doxygen

Bridiro:
# Purpose
Add `Doxyfile` as well as the new `commit-msg` hook, and move Github Actions to our new self-hosted runners.

# Overview
Added `assets/groups.h` as a file used to declare groups. This lets me add `\ingroup` directive to **Doxygen** headers in my files (I did not do it for generated files). This was to organize generation better, to keep both cores in the same website.

# Guidance
The hook and the actions are simply working correctly, nothing to check there.
For Doxygen, try to run it yourself and check the output.
I also added a `README.md` as a "main page" for Doxygen


Doxyfile
```diff
@@ -0,0 +1,2900 @@
...
+# title of most generated pages and in a few other places.
+# The default value is: My Project.
+
+PROJECT_NAME           = "steering-wheel-sw"
+
+# The PROJECT_NUMBER tag can be used to enter a project or revision number. This
+# could be handy for archiving the generated documentation or if some version
+# control system is used.
+
+PROJECT_NUMBER         = "v1.0.0"
```
Tonidotpy: Since it is still in progress it should not be set to version 1.0.0:
```suggestion
PROJECT_NUMBER         = "v0.1.0"
```

README.md
```diff
@@ -0,0 +1,46 @@
...
+    pio test -e tests_common -e tests_cm4 -e tests_cm7
+    ```
+
+## Mantainers
+
+- [Bridi Alessandro](https://github.com/bridiro)
+
+## License
+
+The code is licensed under the **GNU Affero General Public License v3.0**. See the [LICENSE](LICENSE) file for more details.
```
Tonidotpy: No need to specify the LICENSE here since it is automatically displayed by GitHub

assets/groups.h
Tonidotpy, line 1: Not really sure if it is ok or not, I'll let you decide

Tonidotpy (changes requested):

assets/groups.h
Bridiro: In my opinion it's the cleanest solution. I already tried locally to create groups in main files, but modifying generated headers just seemed wrong to me, so I decided that this solution is the best non-destructive one.

Tonidotpy (approved):

PR #16, opened 2026-04-01
Feat/POST

### LINKED ISSUE #8: POST
# Description

The `INIT` state in the **FSM** requires to execute a **POST** (Power on Self-Test). This needs to be implemented in order to check if the board is ok before starting the actual application.

## Implementation Steps

1. Analyze what needs to be tested
2. Implement tests for selected peripherals

## Dependencies

None

## Additional Notes

None

EDITS:
The POST module will be used also to init every module, passing a structure to the FSM containing all hardware-related callbacks.

Bridiro:
# Purpose
The solely purpose of this PR is to add a POST module to both cores.
Right now the POST module only initializes the various modules of the cores, but will be extended for each module added.

# Overview
Simply added **post** module to both cores. They are the same in term of structure and API, but the working is different based on what they should perform.

I had to modify IPC (and consequently inputs) so that the critical-section callback is now typedef'd, to make the post itself clearer.

# Guidance
Try to build the project (running `pio run -e cm4 -e cm7`) and check the modules for correct working (only by watching them for now).

This PR may close #8

### DIFF
--- a/CM4/Core/Inc/steering/post.h
+++ b/CM4/Core/Inc/steering/post.h
@@ -1,0 +1,26 @@
+/*!
+ * \file post.h
+ * \date 2026-04-01
+ * \authors Alessandro Bridi [ale.bridi15@gmail.com]
+ *
+ * \brief This file defines Power-On Self-Test (POST) structures for system diagnostics.
+ */
+
+#ifndef POST_H
+#define POST_H
+
+#include "leds.h"
+#include "ipc.h"
+
+enum PostReturnCode {
+    POST_RC_OK,           /*!< POST completed successfully. */
+    POST_RC_ERROR_MINOR,  /*!< POST encountered a minor error. */
+    POST_RC_ERROR_SEVERE, /*!< POST encountered a severe error. */
+};
+
+struct PostInitData {
+    ipc_critical_section_callback ipc_critical_section; /*!< Callback function required by leds module. */
+    leds_transmit_callback leds_transmit;               /*!< Callback function required by leds module. */
+};
+
+#endif // POST_H
--- a/CM4/Core/Inc/steering/post.h
+++ b/CM4/Core/Inc/steering/post.h
@@ -13,9 +13,8 @@
 #include "ipc.h"
 
 enum PostReturnCode {
-    POST_RC_OK,           /*!< POST completed successfully. */
-    POST_RC_ERROR_MINOR,  /*!< POST encountered a minor error. */
-    POST_RC_ERROR_SEVERE, /*!< POST encountered a severe error. */
+    POST_RC_OK,    /*!< POST completed successfully. */
+    POST_RC_ERROR, /*!< POST encountered an error. */
 };
 
 struct PostInitData {
--- a/CM4/Core/Src/steering/fsm.c
+++ b/CM4/Core/Src/steering/fsm.c
@@ -17,10 +17,9 @@ Functions and types have been generated with prefix "fsm_"
 
 /*** USER CODE BEGIN MACROS ***/
 
-#include "eagletrt-api.h"
 #include "inputs-api.h"
-#include "ipc-api.h"
 #include "leds-api.h"
+#include "post-api.h"
 
 /*** USER CODE END MACROS ***/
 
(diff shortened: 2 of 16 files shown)


CM7/Core/Inc/steering/post.h
```diff
@@ -0,0 +1,20 @@
...
+ * \brief This file defines Power-On Self-Test (POST) structures for system diagnostics.
+ */
+
+#ifndef POST_H
+#define POST_H
+
+enum PostReturnCode {
+    POST_RC_OK,           /*!< POST completed successfully. */
+    POST_RC_ERROR_MINOR,  /*!< POST encountered a minor error. */
+    POST_RC_ERROR_SEVERE, /*!< POST encountered a severe error. */
```
Tonidotpy: What's the meaning of minor and severe error, how do I classify an error as minor or severe?

CM4/Core/Src/steering/fsm.c
```diff
@@ -95,13 +89,9 @@ fsm_state_t fsm_do_init(fsm_state_data_t *data) {
...
-
-    if (inputs_api_init(fsm_data->critical_section_callback, ipc_api_push_event, input_action_noop) != INPUTS_RC_OK) {
-        next_state = FSM_STATE_ERROR;
-    }
+    struct PostInitData *post_init_data = (struct PostInitData *)data;
 
-    if (leds_api_init(fsm_data->leds_transmit) != LEDS_RC_OK) {
+    if (post_api_do_init(post_init_data) != POST_RC_OK) {
         next_state = FSM_STATE_ERROR;
     }
```
Tonidotpy, line 96: What's the point of defining minor and severe error if in both cases you transition the FSM to the ERROR state?

CM4/Core/Src/steering/fsm.c
```diff
@@ -125,9 +115,12 @@ fsm_state_t fsm_do_idle(fsm_state_data_t *data) {
 
     /*** USER CODE BEGIN DO_IDLE ***/
 
-    struct FsmData *fsm_data = (struct FsmData *)data;
-
-    if (inputs_api_poll_for_long_press(fsm_data->get_tick()) != INPUTS_RC_OK) {
+    if (data != NULL) {
+        struct FsmData *fsm_data = (struct FsmData *)data;
+        if (inputs_api_poll_for_long_press(fsm_data->tick) != INPUTS_RC_OK) {
+            next_state = FSM_STATE_ERROR;
```
Tonidotpy, line 121: Does this function always produce an error which is not recoverable (i.e. go to the error state)?

CM4/Core/Inc/steering/post.h
```diff
@@ -0,0 +1,26 @@
...
+#ifndef POST_H
+#define POST_H
+
+#include "leds.h"
+#include "ipc.h"
+
+enum PostReturnCode {
+    POST_RC_OK,           /*!< POST completed successfully. */
+    POST_RC_ERROR_MINOR,  /*!< POST encountered a minor error. */
+    POST_RC_ERROR_SEVERE, /*!< POST encountered a severe error. */
```
Tonidotpy: Not clear what defines a minor and sever error

Tonidotpy (commented):
I have a couple of doubts that I prefer to clear before merging

CM4/Core/Src/steering/fsm.c
Bridiro: the only error that this function can return is if it was unable to handle the possible produced event (e.g. not being able to notify CM7 or not being able to handle it locally, that difficult but still possible). I did it this way but we can still discuss it and decide wether or not to go to ERROR state.

CM4/Core/Src/steering/fsm.c
Bridiro: I actually did it ahead of time for now, because as we know POST will handle peripheral checking at startup, and in the future if a peripheral is not working but it's not essential, additional logic will handle what to do in a better way.

CM4/Core/Inc/steering/post.h
Bridiro: As said in the first comment, right now we don't really have a way to tell and we should discuss this in a meeting with the team, to decide what is recoverable and what should fail the whole board.

PR #17, opened 2026-04-15
Feat/shared data

### LINKED ISSUE #14: feat: data communication between cores
### Description:

As the CM4 is the one reading data from the car, it needs to communicate needed data to CM7.

### Possible implementation:

Similar to how we did for input events, use a double buffer in shared memory, using HSEM to sync reading and writing. Remember that this type of information accept the loss of data, and only needs the most recent.

### Possible issues:

Keeping it clean might be a problem, as it's a feature that feels really tied to hardware.

Bridiro:
# Purpose
To add the skeleton of the structure used to share data between cores.

# Overview
Except a few changes in the whole `ipc` queue (changed because ipc means inter-processor-communication and both the event queue and this are under that name) and other formal changes, only a structure and it's allocation has been added.
The concept is that no locking mechanism is needed as only CM4 writes and CM7 reads, and every operation is considered atomic.

# Guidance
Just check out the structure (and the changes in structure in other modules as well, as well as the flags script and others)


Tonidotpy (approved):
Seems ok to me 👍🏼

PR #24, opened 2026-06-15
Feat: Simulator

### LINKED ISSUE #22: feat: simulator
### Description:

The idea is to have a simulator that runs locally on a computer, that simulates the steering wheel, so that it's easier to develop.

### Possible implementation:

1. Use SDL to simulate how the buffer in the steering wheel behaves
2. Would be better to also simulate the CM4, so another thread or process that communicates with CM7 (maybe one can be enough).
3. Ideally everything would behave exactly like it should on the real hardware.

### Possible issues:

integrate everything cleanly is surely an issue

Bridiro:
# Purpose
Add a "simulator", that is just a windows that opens and runs both FSMs, to simulate the behavior of the whole steering wheel.

# Overview
Right now has been implemented:
- CM4's FSM
- CM7's FSM
- only graphic, inputs, and leds work for the moment

By implementing new features, they should be mirrored in the simulator to not lack behind.

# Guidance
Only changes in the `simulator/` folder were made, because all the code is platform-agnostic. Should work out-of-the-box for MacOS and Linux, but has been tested only on MacOS at the moment.
`tigr.*` files can be skipped as they are copy-pasted from the tigr project. There are some helpers to decode leds duty cycles (to really simulate how they should behave) and some helpers to redefine every symbol from CM4's FSM and POST (conflicting with CM7's).


Tonidotpy:
There is a compilation error and a couple of warnings inside the generated font.

Tonidotpy:
@Bridiro Did you forget to push part of the code? I tried to fix the error manually and it compiles but the font is all ducked up when I run the simulator.

Tonidotpy (commented):
I tried to run static code analysis but it started checking every file, especially the Drivers 😅.
Double check with static code analysis if possible.

For the rest I think it works fine, the only thing it didn't regenerate correctly the font the first time, I had to remove the old font and regenerate it manually.

PR #25, opened 2026-06-16
feat: can communication

### LINKED ISSUE #23: feat: can module and shared memory
### Description:

Right now parameters live in the CM4, and get sent to the CM7 in the queue, but are not written in the shared memory. That's because the idea is to write them to CAN, then update the shared memory structure just from data read from CAN (because ECU relays the parameters we set). This is the reason as of today, IPCData is still "unused".
We have to write a module that uses PAL to receive data and sets the selected ones inside the structure, so that CM7 can display that for us.

### Possible implementation:

1. create the module (probably `communication.c` or something like that)
2. use `PAL` to abstract the peripheral (reception in the `CAN` callback, then synchronous "consuming")
3. Consumed data should update the structure

### Possible issues:

Hopefully none

Bridiro:
# Purpose
Add an abstracted layer over the CAN interface to let the user define standardized packets.
The internally-defined packets `CanCommunicationFrame` holds already-serialized frames + id + length. This helps with consistency and space occupied.
This means that `libcan-sw` comes in just before and after this library.

# Overview
Here is the flow:
- receiving:
    1. init the module with `can_communication_api_init(network, config)`
    2. create a `CanCommunicationFrame` directly inside the CAN interrupt handler (with the received data, of course)
    3. give it to `can_communication_api_add_to_rx(network, frame)`, this copies it as-is into the internal queue
    4. when it's time to process received data, call `can_communication_api_process_rx(network)`, this consumes each message in the queue, passing each frame to a user-defined callback that is passed in by the user at init time (usually, this function uses `libcan-sw` to deserialize and then route with a big switch-case

- sending:
    1. init the module with `can_communication_api_init(network, config)`
    2. create a message using what `libcan-sw` gives you, that is the "easy" API
    3. serialize it into a `CanCommunicationFrame` and give it to `can_communication_api_add_to_tx(network, frame)`
    4. when it's time, process data to send using `can_communication_api_process_tx(network)`. This sends all the already-serialized messages using a user-defined callback passed to the module at init time (usually contains the `HAL_FDCAN_AddMessageToTxQ(...)`.

# Guidance
This is probably in early stages and I'm just looking if what I did might be good or there are better ways.
My reasoning is that, generally without thinking at the exact API, having a medium that is a structure, keeps everything more clean and understandable, letting the user handle `libcan` and its changes over time without actually touching this module.

### DIFF
--- a/CM4/Core/Inc/steering/can-communications.h
+++ b/CM4/Core/Inc/steering/can-communications.h
@@ -54,7 +54,6 @@ enum CanCommunicationReturnCode {
     CAN_COMMUNICATION_RC_NULL_POINTER,          /*!< A required pointer argument was NULL. */
     CAN_COMMUNICATION_RC_INVALID_NETWORK,       /*!< Network ID is out of range or already initialized. */
     CAN_COMMUNICATION_RC_INVALID_LENGTH,        /*!< Frame length is larger than CAN_COMMUNICATIONS_FRAME_DATA_SIZE. */
-    CAN_COMMUNICATION_RC_NOT_INITIALIZED,       /*!< API called on a network whose init never ran. */
     CAN_COMMUNICATION_RC_QUEUE_FULL,            /*!< Target queue has no room. */
     CAN_COMMUNICATION_RC_QUEUE_EMPTY,           /*!< Source queue is empty. */
     CAN_COMMUNICATION_RC_TRANSMISSION_ERROR,    /*!< The user-supplied send callback reported failure. */
@@ -169,7 +168,6 @@ struct CanCommunicationsNetworkState {
     struct PalHandler pal;                          /*!< PAL handler that owns the per-network queues */
     can_communications_send_callback send;          /*!< User-supplied "actually transmit" callback */
     can_communications_receive_callback on_receive; /*!< User-supplied per-frame dispatcher */
-    bool initialized;                               /*!< Whether init has run for this network */
 };
 
 /*!
--- a/CM4/Core/Inc/steering/can-communications.h
+++ b/CM4/Core/Inc/steering/can-communications.h
@@ -10,19 +10,7 @@
  * \details The CAN module wraps a PAL handler per network and presents a
  *     queue-based API. It deliberately stays network-agnostic: frames
  *     travel as raw {id, length, data} tuples in both directions and the
- *     application is responsible for marshalling them through libcan
- *     (the auto-generated can_*_api_serialize_from_id /
- *     can_*_api_deserialize_from_id helpers, one header per network).
- *
- *     libcan therefore lives in the application/dispatch layer, not inside
- *     this module:
- *       - **TX path**: application calls can_*_api_serialize_from_id(...)
- *         to fill the data bytes, then hands the resulting frame to
- *         can_communications_api_send.
- *       - **RX path**: application drains the queue with
- *         can_communications_api_process_rx and runs the matching
- *         can_*_api_deserialize_from_id(...) on the popped frame.
- *
+ *     application is responsible for moving them through libcan.
  *     The module only ever moves bytes; one CAN module covers every
  *     network because libcan keeps each one in its own translation unit.
  */
@@ -41,7 +29,7 @@
  * \brief Maximum payload size of a classic-CAN frame, in bytes.
  *
  * \details FDCAN in classic mode is limited to 8 data bytes. The PAL queues
- *     are sized off the wire-frame buffer that wraps this payload, so any
+ *     are sized off the raw-frame buffer that wraps this payload, so any
  *     bump here automatically widens the internal staging buffers.
  */
 #define CAN_COMMUNICATIONS_FRAME_DATA_SIZE (8U)
@@ -60,15 +48,16 @@
  * \brief Return codes for CAN-communications operations.
  */
 enum CanCommunicationReturnCode {
-    CAN_COMMUNICATION_RC_OK,                 /*!< Operation completed successfully. */
-    CAN_COMMUNICATION_RC_NULL_POINTER,       /*!< A required pointer argument was NULL. */
-    CAN_COMMUNICATION_RC_INVALID_NETWORK,    /*!< Network ID is out of range or already initialized. */
-    CAN_COMMUNICATION_RC_INVALID_LENGTH,     /*!< Frame length is larger than CAN_COMMUNICATIONS_FRAME_DATA_SIZE. */
-    CAN_COMMUNICATION_RC_NOT_INITIALIZED,    /*!< API called on a network whose init never ran. */
-    CAN_COMMUNICATION_RC_QUEUE_FULL,         /*!< Target queue has no room. */
-    CAN_COMMUNICATION_RC_QUEUE_EMPTY,        /*!< Source queue is empty. */
-    CAN_COMMUNICATION_RC_TRANSMISSION_ERROR, /*!< The user-supplied send callback reported failure. */
-    CAN_COMMUNICATION_RC_ERROR,              /*!< Generic failure (PAL/arena/internal). */
+    CAN_COMMUNICATION_RC_OK,                    /*!< Operation completed successfully. */
+    CAN_COMMUNICATION_RC_NULL_POINTER,          /*!< A required pointer argument was NULL. */
+    CAN_COMMUNICATION_RC_INVALID_NETWORK,       /*!< Network ID is out of range or already initialized. */
+    CAN_COMMUNICATION_RC_INVALID_LENGTH,        /*!< Frame length is larger than CAN_COMMUNICATIONS_FRAME_DATA_SIZE. */
+    CAN_COMMUNICATION_RC_NOT_INITIALIZED,       /*!< API called on a network whose init never ran. */
+    CAN_COMMUNICATION_RC_QUEUE_FULL,            /*!< Target queue has no room. */
+    CAN_COMMUNICATION_RC_QUEUE_EMPTY,           /*!< Source queue is empty. */
+    CAN_COMMUNICATION_RC_TRANSMISSION_ERROR,    /*!< The user-supplied send callback reported failure. */
+    CAN_COMMUNICATION_RC_RECEIVE_HANDLER_ERROR, /*!< The user-supplied receive callback reported failure. */
+    CAN_COMMUNICATION_RC_ERROR,                 /*!< Generic failure (PAL/arena/internal). */
 };
 
 /*!
@@ -98,7 +87,7 @@ struct CanCommunicationFrame {
 };
 
 /*!
- * \brief Callback invoked when the module needs to put a frame on the wire.
+ * \brief Callback invoked when the module needs to put a frame on the raw_frame.
  *
  * \details Registered through can_communications_api_init. Wraps the
  *     vendor HAL call (typically HAL_FDCAN_AddMessageToTxFifoQ on the
@@ -113,6 +102,36 @@ struct CanCommunicationFrame {
 typedef enum CanCommunicationReturnCode (*can_communications_send_callback)(const struct CanCommunicationFrame *frame);
 
 /*!
+ * \brief Callback invoked once per frame drained by can_communications_api_process_rx.
+ *
+ * \details The natural home for the libcan deserialise + switch-on-id:
+ *
+ *     \code
+ *     static enum CanCommunicationReturnCode primary_on_receive(const struct CanCommunicationFrame *frame) {
+ *         union CanPrimaryMessages decoded;
+ *         if (can_primary_api_deserialize_from_id(frame->id, frame->data, &decoded) < 0) {
+ *             return CAN_COMMUNICATION_RC_RECEIVE_HANDLER_ERROR;
+ *         }
+ *         switch (frame->id) {
+ *             case CAN_PRIMARY_MESSAGE_FRAME_ID_ECU_STATUS: ...; break;
+ *             ...
+ *         }
+ *         return CAN_COMMUNICATION_RC_OK;
+ *     }
+ *     \endcode
+ *
+ *     Returning anything other than CAN_COMMUNICATION_RC_OK does not stop
+ *     the drain — the rest of the queue is still processed — but the
+ *     final return code of process_rx will surface the error.
+ *
+ * \param[in] frame The frame just popped off the RX queue.
+ *
+ * \retval CAN_COMMUNICATION_RC_OK on success.
+ * \retval CAN_COMMUNICATION_RC_RECEIVE_HANDLER_ERROR if dispatch fails.
+ */
+typedef enum CanCommunicationReturnCode (*can_communications_receive_callback)(const struct CanCommunicationFrame *frame);
+
+/*!
  * \brief Critical-section enter/exit callback type.
  *
  * \details The module's RX queue is pushed from the FDCAN ISR and popped
@@ -124,6 +143,20 @@ typedef enum CanCommunicationReturnCode (*can_communications_send_callback)(cons
 typedef void (*can_communications_critical_section_callback)(void);
 
 /*!
+ * \brief Init-time configuration for one CAN network.
+ *
+ * \details Bundled in a struct (rather than as four positional arguments
+ *     on init) so adding a new wiring point later is a struct extension
+ *     instead of an API break.
+ */
+struct CanCommunicationsNetworkConfig {
+    can_communications_send_callback send;                 /*!< Required: wraps HAL_FDCAN_AddMessageToTxFifoQ */
+    can_communications_receive_callback on_receive;        /*!< Required: dispatch + libcan deserialise */
+    can_communications_critical_section_callback cs_enter; /*!< Optional, may be NULL */
+    can_communications_critical_section_callback cs_exit;  /*!< Optional, may be NULL */
+};
+
+/*!
  * \brief Per-network slot in the module's file-static handler.
  *
  * \details Bundled here (instead of as parallel arrays in the impl) so the
@@ -131,9 +164,10 @@ typedef void (*can_communications_critical_section_callback)(void);
  *     outside can-communications-api.c.
  */
 struct CanCommunicationsNetworkState {
-    struct PalHandler pal;                 /*!< PAL handler that owns the per-network queues */
-    can_communications_send_callback send; /*!< User-supplied "actually transmit" callback */
-    bool initialized;                      /*!< Whether init has run for this network */
+    struct PalHandler pal;                          /*!< PAL handler that owns the per-network queues */
+    can_communications_send_callback send;          /*!< User-supplied "actually transmit" callback */
+    can_communications_receive_callback on_receive; /*!< User-supplied per-frame dispatcher */
+    bool initialized;                               /*!< Whether init has run for this network */
 };
 
 /*!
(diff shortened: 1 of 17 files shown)


CM4/Core/Src/steering/can-communications-router-api.c
Tonidotpy, line 1: To stick with the conventions this file should also have the `api` suffix in theory.

CM4/Core/Inc/steering/can-communications.h
```diff
@@ -0,0 +1,184 @@
...
+
+/*!
+ * \brief Depth of the per-network reception queue.
+ */
+#define CAN_COMMUNICATIONS_RX_QUEUE_CAPACITY (32U)
+
+/*!
+ * \brief Depth of the per-network transmission queue.
+ */
+#define CAN_COMMUNICATIONS_TX_QUEUE_CAPACITY (32U)
```
Tonidotpy, line 47: For this I would use the total number of messages defined inside the canlib, once the library is ready you may change them if needed,

CM4/Core/Inc/steering/post.h
```diff
@@ -19,8 +20,12 @@ enum PostReturnCode {
...
 
 struct PostInitData {
-    leds_transmit_callback leds_transmit;               /*!< Callback function required by leds module. */
-    parameters_on_change_callback parameters_on_change; /*!< Callback fired on every parameter transition. */
+    leds_transmit_callback leds_transmit;                                                       /*!< Callback function required by leds module. */
+    parameters_on_change_callback parameters_on_change;                                         /*!< Callback fired on every parameter transition. */
+    can_communications_send_callback can_send[CAN_COMMUNICATION_NETWORK_COUNT];                 /*!< Array of callbacks for sending CAN frames, one per network. */
+    can_communications_receive_callback can_on_receive[CAN_COMMUNICATION_NETWORK_COUNT];        /*!< Array of callbacks for receiving CAN frames, one per network. */
+    can_communications_critical_section_callback can_cs_enter[CAN_COMMUNICATION_NETWORK_COUNT]; /*!< Array of callbacks for entering critical sections, one per network. Optional, may be NULL. */
+    can_communications_critical_section_callback can_cs_exit[CAN_COMMUNICATION_NETWORK_COUNT];
```
Tonidotpy: Why do you use this if you have the struct?

CM4/Core/Src/fdcan.c
```diff
@@ -250,4 +252,64 @@ void HAL_FDCAN_MspDeInit(FDCAN_HandleTypeDef *fdcanHandle) {
...
+    FDCAN_TxHeaderTypeDef header = { 0 };
+    header.Identifier = frame->id;
+    header.IdType = FDCAN_STANDARD_ID;
+    header.TxFrameType = FDCAN_DATA_FRAME;
+    header.DataLength = (uint32_t)(frame->length << 16U);
+    header.ErrorStateIndicator = FDCAN_ESI_ACTIVE;
+    header.BitRateSwitch = FDCAN_BRS_OFF;
+    header.FDFormat = FDCAN_CLASSIC_CAN;
+    header.TxEventFifoControl = FDCAN_STORE_TX_EVENTS;
+    header.MessageMarker = 0;
```
Tonidotpy: It should be a bit more clear like this
```suggestion
    FDCAN_TxHeaderTypeDef header = {
        .Identifier = frame->id,
        .IdType = FDCAN_STANDARD_ID,
        .TxFrameType = FDCAN_DATA_FRAME,
        .DataLength = (uint32_t)(frame->length << 16U),
        .ErrorStateIndicator = FDCAN_ESI_ACTIVE,
        .BitRateSwitch = FDCAN_BRS_OFF,
        .FDFormat = FDCAN_CLASSIC_CAN,
        .TxEventFifoControl = FDCAN_STORE_TX_EVENTS,
        .MessageMarker = 0
    };
```

CM4/Core/Src/fdcan.c
```diff
@@ -250,4 +252,64 @@ void HAL_FDCAN_MspDeInit(FDCAN_HandleTypeDef *fdcanHandle) {
...
```
Tonidotpy: Same here:
```suggestion
    FDCAN_TxHeaderTypeDef header = {
        .Identifier = frame->id,
        .IdType = FDCAN_STANDARD_ID,
        .TxFrameType = FDCAN_DATA_FRAME,
        .DataLength = (uint32_t)(frame->length << 16U),
        .ErrorStateIndicator = FDCAN_ESI_ACTIVE,
        .BitRateSwitch = FDCAN_BRS_OFF,
        .FDFormat = FDCAN_CLASSIC_CAN,
        .TxEventFifoControl = FDCAN_STORE_TX_EVENTS,
        .MessageMarker = 0
    };
```

CM4/Core/Src/fdcan.c
```diff
@@ -250,4 +252,64 @@ void HAL_FDCAN_MspDeInit(FDCAN_HandleTypeDef *fdcanHandle) {
...
+    if (HAL_FDCAN_AddMessageToTxFifoQ(&hfdcan2, &header, frame->data) != HAL_OK) {
+        return CAN_COMMUNICATION_RC_TRANSMISSION_ERROR;
+    }
+    return CAN_COMMUNICATION_RC_OK;
+}
+
+void HAL_FDCAN_RxFifo0Callback(FDCAN_HandleTypeDef *hfdcan, uint32_t RxFifo0ITs) {
+    if ((RxFifo0ITs & FDCAN_IT_RX_FIFO0_NEW_MESSAGE) != RESET) {
+        FDCAN_RxHeaderTypeDef header = { 0 };
+        struct CanCommunicationFrame msg;
```
Tonidotpy: I'd initialize the struct to make sure it does not have random data.
```suggestion
        struct CanCommunicationFrame msg = { 0 };
```

CM4/Core/Src/fdcan.c
```diff
@@ -250,4 +252,64 @@ void HAL_FDCAN_MspDeInit(FDCAN_HandleTypeDef *fdcanHandle) {
...
+        msg.id = header.Identifier;
+        msg.length = (uint8_t)(header.DataLength >> 16U);
+        can_communications_api_add_to_rx_buffer(hfdcan == &hfdcan1 ? CAN_COMMUNICATION_NETWORK_PRIMARY : CAN_COMMUNICATION_NETWORK_SECONDARY, &msg);
+    }
+}
+
+void HAL_FDCAN_RxFifo1Callback(FDCAN_HandleTypeDef *hfdcan, uint32_t RxFifo1ITs) {
+    if ((RxFifo1ITs & FDCAN_IT_RX_FIFO1_NEW_MESSAGE) != RESET) {
+        FDCAN_RxHeaderTypeDef header = { 0 };
+        struct CanCommunicationFrame msg;
```
Tonidotpy: Same here:
```suggestion
        struct CanCommunicationFrame msg = { 0 };
```

CM4/Core/Src/steering/fsm.c
```diff
@@ -130,6 +131,12 @@ fsm_state_t fsm_do_idle(fsm_state_data_t *data) {
         next_state = FSM_STATE_ERROR;
     }
 
+    EAGLETRT_API_UNUSED(can_communications_api_process_rx(CAN_COMMUNICATION_NETWORK_PRIMARY));
+    EAGLETRT_API_UNUSED(can_communications_api_process_rx(CAN_COMMUNICATION_NETWORK_SECONDARY));
+
+    EAGLETRT_API_UNUSED(can_communications_api_process_tx(CAN_COMMUNICATION_NETWORK_PRIMARY));
+    EAGLETRT_API_UNUSED(can_communications_api_process_tx(CAN_COMMUNICATION_NETWORK_SECONDARY));
```
Tonidotpy, line 139: I'd put a TODO unless you have a valid reason to ignore the return value of those functions.

CM4/Core/Src/steering/can-communications-api.c
Tonidotpy, line 1: I understand now what you meant for the initialization. I think there is a better alternative, you should call the init function one time only from the POST module (since it is the initialization function **of the module**) and inside the init function you can call an internal function of the can communication module to initialize each peripheral separately without the need to check for double initialization, etc....

CM4/Core/Src/steering/can-communications-api.c
```diff
@@ -0,0 +1,346 @@
...
+            prv_pal_send_shims[network],
+            config->cs_enter,
+            config->cs_exit,
+            &handler.arena) != PAL_RC_OK) {
+        return CAN_COMMUNICATION_RC_ERROR;
+    }
+
+    handler.networks[network].initialized = true;
+    return CAN_COMMUNICATION_RC_OK;
+}
```
Tonidotpy: Not good, look at the comment i left on the whole file.
P.s. you are not even checking the initialized variable 😂

CM4/Core/Src/steering/can-communications-api.c
```diff
@@ -0,0 +1,346 @@
...
+ *     Tracked via a static flag rather than a sentinel inside the arena
+ *     struct so we don't depend on arena-allocator's internals.
+ */
+EAGLETRT_STATIC void prv_ensure_arena_ready(void) {
+    EAGLETRT_STATIC bool arena_initialized = false;
+    if (!arena_initialized) {
+        arena_allocator_api_init(&handler.arena);
+        arena_initialized = true;
+    }
+}
```
Tonidotpy: This is kind of ugly and you are using a static variable inside the function.
Also remember that inside unit tests the static attribute is dropped and so it brakes 😄

CM4/Core/Src/steering/can-communications-api.c
```diff
@@ -0,0 +1,346 @@
...
+ * \brief Whether \p network is a valid index into the per-network arrays.
+ *
+ * \param[in] network Network ID to check.
+ *
+ * \retval true  if \p network is in range.
+ * \retval false otherwise.
+ */
+EAGLETRT_STATIC bool prv_network_is_valid(enum CanCommunicationNetwork network) {
+    return network < CAN_COMMUNICATION_NETWORK_COUNT;
+}
```
Tonidotpy: Do you need a function for this?

CM4/Core/Src/steering/can-communications-api.c
```diff
@@ -0,0 +1,346 @@
...
+ *
+ * \details Extending the module to a new network is "add an enum entry +
+ *     add the matching shim + add it here". The size of the array is
+ *     pinned to CAN_COMMUNICATION_NETWORK_COUNT, so the compiler flags any
+ *     mismatch.
+ */
+EAGLETRT_STATIC const pal_send_callback prv_pal_send_shims[CAN_COMMUNICATION_NETWORK_COUNT] = {
+    [CAN_COMMUNICATION_NETWORK_PRIMARY] = prv_pal_send_primary,
+    [CAN_COMMUNICATION_NETWORK_SECONDARY] = prv_pal_send_secondary,
+};
```
Tonidotpy: Why is it called shims??
You can also put this inside the init function since it is used only there.

CM4/Core/Src/steering/can-communications-api.c
```diff
@@ -0,0 +1,346 @@
...
+ * \brief RawFrame-format size pushed through PAL.
+ *
+ * \details Each queued frame is encoded as
+ *     [id (4 bytes, little-endian)][length (1 byte)][data (8 bytes)]
+ *     so PAL can treat the queue as a flat byte stream regardless of the
+ *     CAN frame shape. Bytes past \c length are still copied through
+ *     untouched (they end up undefined on the consumer side, which is what
+ *     CanCommunicationFrame already documents).
+ */
+#define CAN_COMMUNICATIONS_RAW_FRAME_SIZE (CAN_COMMUNICATIONS_RAW_FRAME_DATA_OFFSET + CAN_COMMUNICATIONS_FRAME_DATA_SIZE)
```
Tonidotpy: Not needed, you can just use a variable as the offset and add the size of the variables.

CM4/Core/Inc/steering/can-communications.h
```diff
@@ -0,0 +1,184 @@
...
+#include <stdint.h>
+
+/*!
+ * \brief Maximum payload size of a classic-CAN frame, in bytes.
+ *
+ * \details FDCAN in classic mode is limited to 8 data bytes. The PAL queues
+ *     are sized off the raw-frame buffer that wraps this payload, so any
+ *     bump here automatically widens the internal staging buffers.
+ */
+#define CAN_COMMUNICATIONS_FRAME_DATA_SIZE (8U)
```
Tonidotpy, line 35: This may also be defined by the canlib, not sure tough. For now you can use this.

CM4/Core/Src/steering/can-communications-api.c
```diff
@@ -0,0 +1,346 @@
...
+ * \param[in]  frame Source frame.
+ * \param[out] raw_frame  Destination buffer, at least CAN_COMMUNICATIONS_RAW_FRAME_SIZE bytes long.
+ */
+EAGLETRT_STATIC void prv_encode_frame(const struct CanCommunicationFrame *frame, uint8_t *raw_frame) {
+    raw_frame[0U] = (uint8_t)(frame->id & 0xFFU);
+    raw_frame[1U] = (uint8_t)((frame->id >> 8U) & 0xFFU);
+    raw_frame[2U] = (uint8_t)((frame->id >> 16U) & 0xFFU);
+    raw_frame[3U] = (uint8_t)((frame->id >> 24U) & 0xFFU);
+    raw_frame[CAN_COMMUNICATIONS_RAW_FRAME_LENGTH_OFFSET] = frame->length;
+    (void)memcpy(&raw_frame[CAN_COMMUNICATIONS_RAW_FRAME_DATA_OFFSET], frame->data, CAN_COMMUNICATIONS_FRAME_DATA_SIZE);
```
Tonidotpy: Instead of bit manipulation and macros you can just memcpy everything and calculate the offset:
```suggestion
    uint16_t offset = 0;
    (void)memcpy(raw_frame + offset, &frame->id, sizeof(frame->id));
    offset += sizeof(frame->id);
    (void)memcpy(raw_frame + offset, &frame->length, sizeof(frame->length));
    offset += sizeof(frame->length);
    (void)memcpy(raw_frame + offset, frame->data, CAN_COMMUNICATIONS_FRAME_DATA_SIZE);
```

CM4/Core/Src/steering/can-communications-api.c
```diff
@@ -0,0 +1,346 @@
...
+EAGLETRT_STATIC void prv_decode_frame(const uint8_t *raw_frame, struct CanCommunicationFrame *frame) {
+    frame->id = (uint32_t)raw_frame[0U] |
+                ((uint32_t)raw_frame[1U] << 8U) |
+                ((uint32_t)raw_frame[2U] << 16U) |
+                ((uint32_t)raw_frame[3U] << 24U);
+    frame->length = raw_frame[CAN_COMMUNICATIONS_RAW_FRAME_LENGTH_OFFSET];
+    if (frame->length > CAN_COMMUNICATIONS_FRAME_DATA_SIZE) {
+        frame->length = CAN_COMMUNICATIONS_FRAME_DATA_SIZE;
+    }
+    (void)memcpy(frame->data, &raw_frame[CAN_COMMUNICATIONS_RAW_FRAME_DATA_OFFSET], CAN_COMMUNICATIONS_FRAME_DATA_SIZE);
```
Tonidotpy: For the decode you can do something like this.
```suggestion
    uint16_t offset = 0;
    (void)memcpy(&frame->id, raw_frame + offset, sizeof(frame->id));
    offset += sizeof(frame->id);
    (void)memcpy(&frame->length, raw_frame + offset, sizeof(frame->length));
    offset += sizeof(frame->length);
    (void)memcpy(frame->data, &raw_frame[CAN_COMMUNICATIONS_RAW_FRAME_DATA_OFFSET], CAN_COMMUNICATIONS_FRAME_DATA_SIZE);
    frame->length = EAGLETRT_API_MIN(frame->length, CAN_COMMUNICATIONS_FRAME_DATA_SIZE);
```

CM4/Core/Src/steering/can-communications-api.c
```diff
@@ -0,0 +1,346 @@
...
+enum CanCommunicationReturnCode can_communications_api_process_tx(enum CanCommunicationNetwork network) {
+    if (!prv_network_is_valid(network)) {
+        return CAN_COMMUNICATION_RC_INVALID_NETWORK;
+    }
+    if (!handler.networks[network].initialized) {
+        return CAN_COMMUNICATION_RC_NOT_INITIALIZED;
+    }
+
+    enum CanCommunicationReturnCode result = CAN_COMMUNICATION_RC_OK;
+    while (true) {
```
Tonidotpy: Avoid while true, this function should be called continuously. At most you can use a for loop with a small number of iterations.

CM4/Core/Src/steering/can-communications-api.c
```diff
@@ -0,0 +1,346 @@
...
+    if (!prv_network_is_valid(network)) {
+        return CAN_COMMUNICATION_RC_INVALID_NETWORK;
+    }
+    if (!handler.networks[network].initialized) {
+        return CAN_COMMUNICATION_RC_NOT_INITIALIZED;
+    }
+
+    const can_communications_receive_callback dispatcher = handler.networks[network].on_receive;
+    enum CanCommunicationReturnCode result = CAN_COMMUNICATION_RC_OK;
+    while (true) {
```
Tonidotpy: Same here, no while true.

Tonidotpy (changes requested):
I'll stop reviewing because this module complexity is a little bit high and it can be simplified and improved.
I also didn't check the unit tests, will do in the next review.
Try to simplify the module logic, avoid using too much functions and function pointers whenever possible.

CM4/Core/Src/steering/can-communications-api.c
Bridiro: The `initialized` variable is not meant to be checked here, as I allow re-initialization, but it's only meant to prevent module usage (any other function) to be called on non-initialized networks.

simulator/src/main.c
```diff
@@ -320,6 +321,16 @@ EAGLETRT_STATIC void simulator_seed_ui_snapshot(void) {
...
+enum CanCommunicationReturnCode simulator_can_send_primary(const struct CanCommunicationFrame *frame) {
+    (void)frame;
+    return CAN_COMMUNICATION_RC_OK;
+}
+
+enum CanCommunicationReturnCode simulator_can_send_secondary(const struct CanCommunicationFrame *frame) {
+    (void)frame;
+    return CAN_COMMUNICATION_RC_OK;
+}
+
```
Tonidotpy, line 333: Remember to use `EAGLETRT_API_UNUSED` instead of a cast to void

CM4/Core/Src/steering/can-communications-api.c
```diff
@@ -0,0 +1,244 @@
...
+    }
+    return result;
+}
+
+enum CanCommunicationReturnCode can_communications_api_process_rx(enum CanCommunicationNetwork network) {
+    if (network >= CAN_COMMUNICATION_NETWORK_COUNT) {
+        return CAN_COMMUNICATION_RC_INVALID_NETWORK;
+    }
+
+    const can_communications_receive_callback dispatcher = handler.networks[network].on_receive;
```
Tonidotpy: Since it is a function callback I'd rather call it `dispatch` or `handle_received_message` or something like that.

CM4/Core/Src/steering/can-communications-api.c
```diff
@@ -0,0 +1,244 @@
...
+            break;
+        }
+        if (return_code == PAL_RC_IO_ERROR) {
+            /* User send callback reported failure on this frame; keep draining. */
+            result = CAN_COMMUNICATION_RC_TRANSMISSION_ERROR;
+            continue;
+        }
+        if (return_code != PAL_RC_OK) {
+            return CAN_COMMUNICATION_RC_ERROR;
+        }
```
Tonidotpy, line 215: I think you can just use an `else` instead of `continue`
```suggestion
        if (return_code == PAL_RC_IO_ERROR) {
            /* User send callback reported failure on this frame; keep draining. */
            result = CAN_COMMUNICATION_RC_TRANSMISSION_ERROR;
        }
        else if (return_code != PAL_RC_OK) {
            return CAN_COMMUNICATION_RC_ERROR;
        }
```

Tonidotpy (changes requested):
Just a couple of small changes.

Tonidotpy (approved):
Seems fine to me for now, we'll see when the canlib is ready and we actually have something to test it on.

